PR #2, opened 2015-04-02
Fixes failing specs

purbon:
This PR make all specs to pass in a more consistent way as it was before, now all are green also in linux environments as our beloved jenkins.

The main problem was the time needed to kill the thread used to run the plugin within the test example, in our test environment it got kill slightly after 3 seconds passed (the desired time frame), so at the end this end up on not having the desired number of events.

To "workaround" this situation for now I propose to call the teardown method just after the sleep call, so the scheduler is off, and not rely on  thread.kill to stop the pipeline properly and on time.

I also added the generated logs to the gitignore file.

So what do you think @talevy ?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 Gemfile.lock
 .bundle
 vendor
+derby.log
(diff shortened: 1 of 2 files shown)


talevy:
I agree. your logic holds. thanks for the correction!

talevy:
LGTM

PR #3, opened 2015-04-10
Clean up ruby-maven dependency

purbon:
cleanup the usage of ruby maven as it's not really used here, and we also aim to not used it anymore in favor of having jars loaded within the gems.

Related to https://github.com/elastic/logstash/issues/2954


talevy:
LGTM

PR #12, opened 2015-07-07
add persistence

talevy:


lib/logstash/inputs/jdbc.rb
```diff
@@ -59,11 +60,27 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # for example: "* * * * *" (execute query every minute, on the minute)
   config :schedule, :validate => :string
 
+  # Path to file with last run time
+  config :last_run_metadata_path, :validate => :string, :default => "#{ENV['HOME']}/.logstash_jdbc_last_run"
```
jordansissel: this should be `:validate => :path`

lib/logstash/inputs/jdbc.rb
```diff
@@ -82,10 +99,19 @@ def teardown
     if @scheduler
       @scheduler.stop
     end
+
+    # update state file for next run
+    if @record_last_run
+      File.open(@last_run_metadata_path, 'w') do |f|
```
jordansissel: This could be shortened with File.write -

```ruby
File.write(@last_run_metadata_path, YAML.dump(@sql_last_start))`
```

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
 
     @database.drop_table(:test_table)
   end
+
+  context "persistence" do
+
+    it "should respect last run metadata" do
```
jordansissel: we can break up these into `let` statements:

```
let(:settings) { { "statement" => .... } }
let(:plugin) { LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings)) }
let(:queue) { Queue.new }

before do
  plugin.register
end
```

lib/logstash/inputs/jdbc.rb
talevy: but that would rely on the file existing beforehand

lib/logstash/inputs/jdbc.rb
talevy: oooo shiny. didnt know!

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
...
+        "statement" => "SELECT * FROM SYSIBM.SYSDUMMY1",
+        "last_run_metadata_path" => Stud::Temporary.pathname
+      }
+      plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
+      plugin.register
+      q = Queue.new
+      plugin.run(q)
+      plugin.teardown
+
+      insist { File.exists?(settings["last_run_metadata_path"]) }
```
jordansissel: In the future, let's use expect(). I'd love remove insist from our codebase now that rspec 3's expect() is so nice :)

```ruby
expect(File).to(be_exists(settings["last_run_metadata_path"]))
```

'be_exists' calls File.exists?(...) and expects it to return a truthy value.

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
...
+      last_run = YAML.load(File.read(settings["last_run_metadata_path"]))
+
+      plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
```
jordansissel: let's move this section to a new 'it' clause if we're testing different things?

On the other hand, if you're testing the behavior of two different runs, I think it makes sense to put the "previous run" in a before clause:

```ruby
before do
  # do first run
  # plugin.teardown, etc...
end

it "..." do
  # Make sure the 2nd run respects the last_run setting?
end
```

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
...
+    end
+
+    it "should ignore last run metadata if :clean_run set to true" do
+      settings = {
+        "statement" => "SELECT * FROM SYSIBM.SYSDUMMY1",
+        "last_run_metadata_path" => Stud::Temporary.pathname,
+        "clean_run" => true
+      }
+
+      File.open(settings["last_run_metadata_path"], "w") { |f| f.write(YAML.dump(Time.at(1).utc)) }
```
jordansissel: similar comments here; move variable-like things into `let` statements.

For this yaml writing, I'm not sure. It feels like it's touching too much of the internal code. Maybe it's ok for now, not sure. Feels like the right way is to wrap the metadata load/store with a module that you invoke directly instead of doing this file writing + yaml dump in the test. Thoughts?

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
...
+
+      plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
+      plugin.register
+
+      insist { plugin.instance_variable_get("@sql_last_start") } == Time.at(0).utc
```
jordansissel: The lines above writes `Time.at(1).utc` and you'er expecting it to equal `Time.at(0).utc` ? This seems... odd?

jordansissel: Oh nevermind, I see this test uses `clean_run`

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
...
+      settings = {
+        "statement" => "SELECT * FROM SYSIBM.SYSDUMMY1",
+        "last_run_metadata_path" => Stud::Temporary.pathname,
+        "record_last_run" => false
+      }
+
+      plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
+      plugin.register
+
+      insist { File.exists?(settings["last_run_metadata_path"]) }
```
jordansissel: This doesn't check anything (insist's block return value is not used). Insist has a `reject` to do the opposite. Also this should be acting on the `File` -

```ruby
reject { File }.exists?(settings["last_run_metadata_path"])
```

Then again, we want to move to using rspec 3's expect(), so this is preferred:

```ruby
expect(File).to(be_exists?(settings["last_run_metadata_path"]))
```

(I think this is preferred, anyway? There may be a different way to express it that doesn't read as awkwardly as "be exists"

spec/inputs/jdbc_spec.rb
```diff
@@ -113,4 +114,59 @@
...
+
+      plugin.teardown
```
jordansissel: if you just need teardown at the end of most things, you can put it in an after clause:

```ruby
after do
  plugin.teardown
end
```

lib/logstash/inputs/jdbc.rb
jordansissel: hmm.. yes, neeevermind then. :)

talevy: I thought so too, and then I got a configerror and realized

spec/inputs/jdbc_spec.rb
talevy: it would be awesome if we could standardize a way between plugins. should we add such a module to logstash-core? (unless it already exists and I missed it)

spec/inputs/jdbc_spec.rb
talevy: hmm, some plugins depend on teardown before done doing assertions. specifically just for scheduling since that spawns a new thread. but I may be able to organize it so the other tests can re-use this

spec/inputs/jdbc_spec.rb
talevy: agreed. I should really start doing this. will push out a new fix PR with an update for this and other cleanups for the rspecs

spec/inputs/jdbc_spec.rb
talevy: my pleasure, will update this

spec/inputs/jdbc_spec.rb
talevy: I was hoping to use some of these across tests, but `:settings` is different in almost every `it` block.

PR #13, opened 2015-07-07
add jdbc fetch size param to queries

talevy:
This means slightly different things depending on the vendor and its JDBC driver implementation. Still collecting proven updated information about each vendor's implementation so this can be documented further

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -42,6 +42,15 @@ require "yaml" # persistence
 # `clean_run` is set to true, this value will be ignored and `sql_last_start` will be
 # set to Jan 1, 1970, as if no query has ever been executed.
 #
+# ==== Dealing With Large Result-sets
+#
+# Many JDBC drivers use the `fetch_size` parameter to limit how many
+# results are pre-fetched at a time from the cursor into the client's cache
+# before retrieving more results from the result-set. This is configured in
+# this plugin using the `jdbc_fetch_size` configuration option. No fetch size
+# is set by default in this plugin, so the specific driver's default size will 
+# be used.
+#
 # ==== Usage:
 #
 # Here is an example of setting up the plugin to fetch data from a MySQL database.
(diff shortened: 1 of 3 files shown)


spec/inputs/jdbc_spec.rb
```diff
@@ -260,4 +260,34 @@
...
+      end
+
+      plugin.register
+    end
+
+    after do
+      plugin.teardown
+    end
+
+    it "should fetch all rows" do
```
suyograo: if `jdbc_fetch_size = 1`, why are we validating it brought back all results?

talevy: I am mainly testing that the library doesn't throw an exception when this parameter is set. fetch_size is defaulted to `10` in some clients. this is not how much of the result set is returned... but how much is cached before the cursor is moved and another batch is fetched from the database. It is more for keeping memory footprint low while trying to retrieve a _really_ large result.

lib/logstash/inputs/jdbc.rb
```diff
@@ -42,6 +42,14 @@
 # `clean_run` is set to true, this value will be ignored and `sql_last_start` will be
 # set to Jan 1, 1970, as if no query has ever been executed.
 #
+# ==== Dealing With Large Result-sets
+#
+# Many JDBC drivers use the `fetch_size` parameter to limit how many results are
+# loaded at a time in the client's cache before retrieving more results from the result
```
suyograo: how about? > parameter to limit how many results are pre-fetched at a time from the cursor into the client's cache

talevy: cleaner, yup! will do.

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -45,6 +45,9 @@ def setup_jdbc_config
     # JDBC page size
     config :jdbc_page_size, :validate => :number, :default => 100000
 
+    # JDBC fetch size
```
suyograo: can we also add the default here is the driver's default

suyograo:
minor doc comment on the config, merge after that

PR #14, opened 2015-07-07
rspec updates and cleanup

talevy:
In response to a lapse in previous reviews :) here goes a PR for updating the tests to remove `insist` and try and be more DRY.

### DIFF
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -1,172 +1,228 @@
 require "logstash/devutils/rspec/spec_helper"
 require "logstash/inputs/jdbc"
 require "jdbc/derby"
+require "sequel"
+require "sequel/adapters/jdbc"
 require "timecop"
 require "stud/temporary"
 
 
 describe "jdbc" do
   let(:mixin_settings) { {"jdbc_driver_class" => "org.apache.derby.jdbc.EmbeddedDriver", "jdbc_connection_string" => "jdbc:derby:memory:testdb;create=true"} }
+  let(:settings) { {} }
+  let(:plugin) { LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings)) }
+  let(:queue) { Queue.new }
+  let (:db) do
+    Sequel.connect(mixin_settings['jdbc_connection_string'], :user=> nil, :password=> nil)
+  end
 
   before :each do
     Jdbc::Derby.load_driver
-  end
-
-  it "should register and tear down" do
-    settings = {"statement" => "SELECT 1 as col1 FROM SYSIBM.SYSDUMMY1"}
-    plugin = LogStash::Plugin.lookup("input", "jdbc").new(mixin_settings.merge(settings))
-    expect { plugin.register }.to_not raise_error
-    expect { plugin.teardown }.to_not raise_error
-  end
-
-  it "should retrieve params correctly from Event" do
-    settings = {"statement" => "SELECT :num_param as num_param FROM SYSIBM.SYSDUMMY1", "parameters" => {"num_param" => 10} }
-    plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
-    plugin.register
-    q = Queue.new
-    plugin.run(q)
-    insist { q.size } == 1
-    insist { q.pop['num_param'] } == settings['parameters']['num_param']
-    plugin.teardown
-  end
-
-  it "should properly schedule" do
-    settings = {"statement" => "SELECT 1 as num_param FROM SYSIBM.SYSDUMMY1", "schedule" => "* * * * * UTC"}
-    plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
-    plugin.register
-    q = Queue.new
-    Timecop.travel(Time.new(2000))
-    Timecop.scale(60)
-    runner = Thread.new do
-      plugin.run(q)
-    end
-    sleep 3
-    plugin.teardown
-    runner.kill
-    runner.join
-    insist { q.size } == 2
-    Timecop.return
-  end
-
-  it "should appropriately page table" do
-    require "sequel"
-    require "sequel/adapters/jdbc"
-    Jdbc::Derby.load_driver
-    @database = Sequel.connect(mixin_settings['jdbc_connection_string'], :user=> nil, :password=> nil)
-    @database.create_table :test_table do
+    db.create_table :test_table do
       DateTime :created_at
       Integer :num
     end
-    test_table = @database[:test_table]
-    settings = {"statement" => "SELECT * from test_table",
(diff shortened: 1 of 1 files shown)


talevy:
still hoping to figure out a way to instantiate more `let` variables across the test cases. I am not completely sure how to do this since plugin instantiation depends on specific settings which are unique to almost all the tests

spec/inputs/jdbc_spec.rb
```diff
@@ -1,12 +1,15 @@
...
 require "jdbc/derby"
+require "sequel"
+require "sequel/adapters/jdbc"
 require "timecop"
 require "stud/temporary"
 
 
 describe "jdbc" do
   let(:mixin_settings) { {"jdbc_driver_class" => "org.apache.derby.jdbc.EmbeddedDriver", "jdbc_connection_string" => "jdbc:derby:memory:testdb;create=true"} }
+  let(:q) { Queue.new }
```
purbon: the name of this variable could be more explanatory :+1: , why not ```queue```? :smile:

purbon:
@talevy let variables are just that variables, who are so nice to be reset all the time, like monoid I guess is call in FP.

Keep in mind you can have multiple describe and context, then you can reuse or redefine any let variable created, inheritance apply between the several context.

spec/inputs/jdbc_spec.rb
```diff
@@ -23,10 +26,9 @@
     settings = {"statement" => "SELECT :num_param as num_param FROM SYSIBM.SYSDUMMY1", "parameters" => {"num_param" => 10} }
     plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
     plugin.register
-    q = Queue.new
     plugin.run(q)
-    insist { q.size } == 1
-    insist { q.pop['num_param'] } == settings['parameters']['num_param']
+    expect(q.size).to eq(1)
```
purbon: Why are you checking for the queue size, if you aim to check for the parameters? Don't understand why it's need.

purbon:
@talevy every time I see something like:

```
settings = {"statement" => "SELECT 1 as num_param FROM SYSIBM.SYSDUMMY1", "schedule" => "* * * * * UTC"}
plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
plugin.register
```

I think you're really facing with another test context, why not having something like:

```
let(:settings) {}
subject(:plugin) {}
...
   before do
      plugin.register
   end
```

then in each context you can redefine the settings to the specific ones.

spec/inputs/jdbc_spec.rb
```diff
@@ -164,7 +163,7 @@
       plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
       plugin.register
 
-      insist { File.exists?(settings["last_run_metadata_path"]) }
+      expect(File).not_to(be_exists(settings["last_run_metadata_path"]))
```
purbon: avoid writing so many parentheses as in here, it makes reading this like complicated. Use something like in https://www.relishapp.com/rspec/rspec-expectations/v/3-3/docs/built-in-matchers/exist-matcher

purbon:
Avoid having test like: https://github.com/logstash-plugins/logstash-input-jdbc/pull/14/files#diff-9c9936a0fd95805a5aa45981f93241c0L15 . You are either testing register or teardown.

purbon:
:dancers: for removing insist!! thanks!

spec/inputs/jdbc_spec.rb
talevy: thanks! I like that

cc @jordansissel

spec/inputs/jdbc_spec.rb
```diff
@@ -1,172 +1,208 @@
...
-    settings = {"statement" => "SELECT :num_param as num_param FROM SYSIBM.SYSDUMMY1", "parameters" => {"num_param" => 10} }
-    plugin = LogStash::Inputs::Jdbc.new(mixin_settings.merge(settings))
-    plugin.register
-    q = Queue.new
-    plugin.run(q)
-    insist { q.size } == 1
-    insist { q.pop['num_param'] } == settings['parameters']['num_param']
-    plugin.teardown
-  end
+    it "should register without raising exception" do
```
talevy: should I remove these tests? kind of awkward.

spec/inputs/jdbc_spec.rb
```diff
@@ -1,172 +1,208 @@
...
-    runner = Thread.new do
-      plugin.run(q)
-    end
-    sleep 3
-    plugin.teardown
-    runner.kill
-    runner.join
-    insist { q.size } == 2
-    Timecop.return
+    it "should teardown without raising exception" do
```
talevy: should I remove these tests? kind of awkward.

spec/inputs/jdbc_spec.rb
```diff
@@ -1,172 +1,208 @@
...
 
-  it "should successfully iterate table with respect to field values" do
-    require "sequel"
-    require "sequel/adapters/jdbc"
-    Jdbc::Derby.load_driver
-    @database = Sequel.connect(mixin_settings['jdbc_connection_string'], :user=> nil, :password=> nil)
-    @database.create_table :test_table do
-      DateTime :created_at
-      Integer :num
+  context "with specific database" do
```
talevy: not sure how to clean up this context. should I split this out into its own `describe` block so that I can declare new `settings` in the same manner as the other tests using `let`?

spec/inputs/jdbc_spec.rb
```diff
@@ -1,172 +1,225 @@
...
-    end
-    sleep 3
-    plugin.teardown
-    runner.kill
-    runner.join
-    insist { q.size } == 2
-    Timecop.return
+    it "should register without raising exception" do
+      expect { plugin.register }.to_not raise_error
+      plugin.teardown
```
purbon: For this, I guess you can use an 

```
after(:each) do
      plugin.teardown
end
```

what do you think?

spec/inputs/jdbc_spec.rb
```diff
@@ -1,172 +1,225 @@
...
+  end
 
-    plugin.run(q)
-    plugin.teardown
+  context "when scheduling" do
+    let(:settings) { {"statement" => "SELECT 1 as num_param FROM SYSIBM.SYSDUMMY1", "schedule" => "* * * * * UTC"} }
 
-    insist { q.size } == NUM_ROWS
+    it "should properly schedule" do
+      plugin.register
```
purbon: why not having this prep and post work in before and after hooks? what do you think?

purbon:
@talevy just left you a very small comment, for the rest this looks very nice. Good work man!

spec/inputs/jdbc_spec.rb
talevy: updated. I will leave rest as is because of timing concerns

talevy:
@purbon thanks for the final comments, updated!

PR #19, opened 2015-07-14
add funcionality to read a sql statement from a file, if path provided

talevy:


lib/logstash/inputs/jdbc.rb
```diff
@@ -81,6 +83,9 @@ def register
     elsif File.exists?(@last_run_metadata_path)
       @sql_last_start = YAML.load(File.read(@last_run_metadata_path))
     end
+
+    # load statement from file, if filepath passed.
+    @statement = File.read(@statement) if File.exists?(@statement)
```
suyograo: can we add an explicit `statement_file` config to take the configuration from a SQL file? Much cleaner IMO

talevy:
@suyograo updated with `statement_filepath`

lib/logstash/inputs/jdbc.rb
```diff
@@ -38,6 +39,14 @@
 #     schedule => "* * * * *"
 #   }
 # }
+#
+# #### statement vs statement_filepath
```
suyograo: `====` instead of `####` for asciidoc :)

suyograo:
LGTM

PR #20, opened 2015-07-14
fix asciidoc header syntax

talevy:
Logstash documentation is generated using asciidoc, not markdown. This change attempts to fix the syntax towards asciidoc


suyograo:
LGTM

PR #23, opened 2015-07-14
Add more description to docs

suyograo:
Removed blob about mixins and moved usage up on top

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -4,25 +4,11 @@ require "logstash/namespace"
 require "logstash/plugin_mixins/jdbc"
 require "yaml" # persistence
 
-# INFORMATION
-#
-# This plugin was created as a way to iteratively ingest any database
-# with a JDBC interface into Logstash.
-#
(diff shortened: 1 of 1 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -42,12 +28,22 @@
...
+# The file option only supports one SQL statement. The plugin will only accept one of the options.
+# It cannot read a statement from a file as well as from the `statement` configuration parameter.
+#
+# ==== Predefined Parameters
+#
+# Some parameters are built-in and can be used from within your queries.
+# Here is the list:
+#
+# |==========================================================
+# |sql_last_start |The time the last query executed in plugin
```
jordansissel: What if we've never run the query? What's the value of sql_last_start?

suyograo: @talevy I did not understand this too? Can you elaborate (in the docs too)

talevy: yup. will clarify this. `sql_last_start` is set to the start of time before a query is executed

PR #27, opened 2015-07-16
Add more descriptive help for Driver Library Errors

### LINKED ISSUE #17: Add check for driver in classpath
It is a requirement to have the JDBC driver libraries in the classpath. We should have a check to be explicit of a driver missing from the classpath. instead of "Not Found" error.

talevy:
When someone does not configure the driver library correctly,
a more helpful message would be great. This is a step in that
direction.

Fixes #17.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -19,6 +19,8 @@ module LogStash::PluginMixins::Jdbc
   public
   def setup_jdbc_config
     # JDBC driver library path to third party driver library.
+    #
+    # If not provided, Plugin will look for the driver class in the Logstash Java classpath.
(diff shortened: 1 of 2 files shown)


lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -19,6 +19,8 @@ def self.included(base)
   public
   def setup_jdbc_config
     # JDBC driver library path to third party driver library.
+    #
+    # If not provided, Plugin will look driver class in the Logstash Java classpath.
```
suyograo: typo. look > look for the

talevy: updated. thanks!

suyograo:
LGTM

PR #30, opened 2015-07-20
Configuration Change and Sore Data Improve

pafq:
Changes:

   Configuration  organized
   Support to Store Persist Data from the Query

Later, will build a filter plugin that use the same Configuration Structure and will allow to make multiple queries.
(https://discuss.elastic.co/t/how-to-get-jdbc-data-from-multiple-queries/25631)


electrical:
Since this change is not backwards compatible it would require a major version bump.

talevy:
@pafq 

I like some of the additions being made, but I am a little unclear on the benefit of the additional `Logstash::Inputs::Statement` abstraction.

I also like the idea of a configurable persistence option, but it seems that `File` is still the only option.

would you mind sharing an example of how you are using this?

pafq:
@talevy 

First I want to say that I start looking to Ruby last week, so my code can be a little messy and not optimized. I usualy program in Java so some concepts could be wrong when apply to Ruby.

The Logstash::Inputs::Statement abstraction it's a kind of organization and if you saw this new plugin I made (https://github.com/elastic/logstash/issues/3633) it is clear the main ideia. In the filter plugin, you can configure a Statement inside another Statement, so is usefull to have the Statement abstraction.

I try to use the same configuration format for the Input and Filter Plugin. Eventually the Statement and Connection can be shared between both plugins.

I didn't understand very well your question about persistence option. Do you want examples for another persistence option? Or you want to know what kind of information I store in the File?

For another examples of persistence option, instead of a File we can use REDIS to Store the data. At this moment I only use the FILE (I'm only evaluating LogStash to see if respond to my company needs).

If your question was about the kind of information I store in the File then, I have some DB Table that don't have any kind of Date Field to apply the initial sql_last_start filter, so I need to store the last id (this is a Primary Key and the Query is made with a Order by Id Asc) I collected and then make a filter with this ID.

guyboertje:
I am closing this.

Much has changed since this was submitted. There is a jdbc streaming filter and shortly there will be a jdbc static filter. Each of these offers slightly different ways of enhancing an event retrieved by a single query in this input.

This input will soon be merged with the above two filters in a Universal Plugin, one gem that can contain multiple LS plugins. At that point we will share the abstractions between all jdbc plugins.

PR #53, opened 2015-09-11
update to new shutdown api

### LINKED ISSUE #52: refactor shutdown sequence to use plugin.stop
Change shutdown sequence to be triggered by plugin.stop instead of ShutdownSignal exception.
Also remove any calls to: shutdown, finished, finished?, running? or terminating?
This depends on https://github.com/elastic/logstash/pull/3895

talevy:
Fixes #52.

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -163,13 +163,14 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
       @scheduler.cron @schedule do
         execute_query(queue)
       end
+
       @scheduler.join
     else
       execute_query(queue)
     end
   end # def run
 
-  def close
+  def stop
     @scheduler.stop if @scheduler
 
     # update state file for next run
@@ -178,7 +179,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
     end
 
     close_jdbc_connection
-  end # def close
+  end
 
   private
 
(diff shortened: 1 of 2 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -163,22 +163,27 @@ def run(queue)
...
       @scheduler.join
     else
       execute_query(queue)
     end
   end # def run
 
-  def teardown
+  def stop
+    super
     @scheduler.stop if @scheduler
```
purbon: In a similar situation at https://github.com/logstash-plugins/logstash-input-neo4j/pull/4/files#diff-6ec72c1ec89ef0f4ba329099faaa459fR60 I used the shutdown method, do you know what works better?

lib/logstash/inputs/jdbc.rb
```diff
@@ -163,22 +163,27 @@ def run(queue)
...
+  end
 
+  def close
```
purbon: why not redefining the stop method here?

lib/logstash/inputs/jdbc.rb
purbon: super would need to be removed here.

purbon:
@talevy left you some comments :+1:

talevy:
@purbon updated!

lib/logstash/inputs/jdbc.rb
```diff
@@ -186,6 +187,7 @@ def execute_query(queue)
     # update default parameters
     @parameters['sql_last_start'] = @sql_last_start
     execute_statement(@statement, @parameters) do |row|
+      break if stop?
```
purbon: would this not be better at the end of the loop? so it checks before actually getting to execute the statement? does this makes sense?

purbon:
LGTM, test pass

PR #64, opened 2015-10-21
add ability to configure timeout and other opts

### LINKED ISSUE #62: Make timeouts configurable
Currently, the timeouts (I think?) are hard-coded within Sequel to be 5 seconds. It could be useful to have this be configurable.

talevy:
Sequel supports various vendor-specific and connection-pool-specific
options that were not exposed to the plugin. This change introduces
a dedicated new configuration option to set a timeout and also a
catch-all configuration option for any other options that Sequel
supports upon opening a new database connection.

Fixes #62.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -57,6 +57,37 @@ module LogStash::PluginMixins::Jdbc
     # Connection pool configuration.
     # How often to validate a connection (in seconds)
     config :jdbc_validation_timeout, :validate => :number, :default => 3600
+
+    # Connection pool configuration.
+    # The amount of seconds to wait to acquire a connection before raising a PoolTimeoutError (default 5)
+    config :jdbc_pool_timeout, :validate => :number, :default => 5
+
+    # General/Vendor-specific Sequel configuration options.
+    #
+    # An example of an optional connection pool configuration
+    #    max_connections - The maximum number of connections the connection pool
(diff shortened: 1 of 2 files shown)


spec/inputs/jdbc_spec.rb
```diff
@@ -5,6 +5,7 @@
 require "sequel/adapters/jdbc"
 require "timecop"
 require "stud/temporary"
+require "rspec/wait"
```
talevy: unused, to be removed

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -76,7 +99,7 @@ def prepare_jdbc_connection
                 end
       raise LogStash::ConfigurationError, "#{e}. #{message}"
     end
-    @database = Sequel.connect(@jdbc_connection_string, :user=> @jdbc_user, :password=>  @jdbc_password.nil? ? nil : @jdbc_password.value)
+    @database = jdbc_connect()
```
suyograo: does this log message when timeout was reached and connection failed?

suyograo:
minor comment, LGTM otherwise

lib/logstash/plugin_mixins/jdbc.rb
talevy: updated to log error when failed to connect because of timeout related reasons, as well as others.

talevy:
updated to reflect comments

PR #69, opened 2015-10-22
pass password as string to Sequel

talevy:
this issue was found with using Logstash's Password class to wrap the password string.

ref: https://github.com/logstash-plugins/logstash-input-jdbc/issues/62#issuecomment-150103578

this commit extracts the string value of the password and passes that to Sequel.


suyograo:
LGTM

PR #70, opened 2015-10-22
prevents multiple queries from being run at the same time.

talevy:
Currently, if a query is still running while the scheduler
triggers a new query to be executed, the new query will be launched.
This means that two queries will be running at the same time, this can
cause consistency issues. By setting the max_work_threads to `1`, all
newly triggered queries will be added to a work queue within Rufus and
launched immedietly proceeding the completion of the existing pending
job.


suyograo:
LGTM

PR #74, opened 2015-10-29
Bump to 2.0.4 and changelog

eriktrom:


eriktrom:
I realize i broke with convention by linking to the issue in the  CHANGELOG let me know if you'd like that changed. I've seen it elsewhere and enjoy so figured I'd share the idea.

(comment crosslink for those stuck w/ issue that sparked this version bump https://github.com/elastic/logstash/issues/3785#issuecomment-152104199)

suyograo:
Thanks for the patch @eriktrom and i really like your idea of linking to the issue. We'll continue that :)

eriktrom:
@suyograo - sounds good - one thing i noticed when linking by hand is that i did the old duplicate line line and almost forgot to edit the source URL of the second line - i caught it but perhaps a little script could be of service here. 

If that's of any interest, give me a :+1: and I'll track you down on slack/irc next week sometime and maybe we can hash something out :)

PR #83, opened 2015-11-16
Provide a fix for RubyTime conversion

### LINKED ISSUE #77: RubyTime object is not Logstash::Timestamp
It appears that the JDBC plugin is grabbing timestamps correctly, but it's only converting them to `RubyTime` objects, which is not the same thing as a `Logstash::Timestamp`.  This conversion should be possible without resorting to a `ruby{}` filter.

See: https://discuss.elastic.co/t/logstash-date-filter-question/33663

purbon:
Fix #77 by adding some decoration/conversion when need.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 2.0.5
+  - [#77](https://github.com/logstash-plugins/logstash-input-jdbc/issues/77) Time represented as RubyTime and not as Logstash::Timestamp
+
 ## 2.0.4
   - [#70](https://github.com/logstash-plugins/logstash-input-jdbc/pull/70) prevents multiple queries from being run at the same time
   - [#69](https://github.com/logstash-plugins/logstash-input-jdbc/pull/69) pass password as string to Sequel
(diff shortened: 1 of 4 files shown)


purbon:
@talevy happy if you can give me your thoughts on this.

suyograo:
nice!

jordansissel:
Code LGTM; haven't tested. Nice refactoring along with the fix! :)

suyograo:
@purbon curious if there is a performance hit of this? going through every field and checking if its a time value? 

Also, on a related note, sequel exposes this option `:convert_types`:

> If set to false, does not attempt to convert some Java types to ruby types. Setting to false roughly doubles performance when selecting large numbers of rows. Note that you can't provide this option inside the connection string (as that is passed directly to JDBC), you have to pass it as a separate option.

See http://sequel.jeremyevans.net/rdoc/files/doc/opening_databases_rdoc.html#label-jdbc+

Should we explore this?

purbon:
@suyograo no doubt we should explore performance on this.

re: not changing previous looping behaviour to check for time, just adding a conversion when need ( I expect not a lot of impact as it was done before just to convert keys to strings).

suyograo:
+1, I'll open a new issue for using `:convert_types`

jordansissel:
> curious if there is a performance hit of this?

We could also take a similar approach to Elasticsearch's dynamic mapping - cache previously-unknown column/field names by type. Should be faster than checking each column/field every event, but i'm open to having it be explicit (even if just for a little while) to start with.

suyograo:
+1 on merging this for a first iteration. We can always make it faster :)

talevy:
@jordansissel PR has build check, no need to test!

LGTM

PR #86, opened 2015-11-19
Enable more the option to load more than one jar as part of JDBC drivers

### LINKED ISSUE #85: jdbc_driver_library parameter should support array value
Sometimes, as is the case with Teradata, the JDBC driver libraries may consist of more than one jar, but the jdbc_driver_library parameter only supports a single argument.

```
terajdbc4.jar 
tdgssconfig.jar
```

Can this parameter be changed to support an array value?

purbon:
Hi, as stated in #85, in some cases you need to load more than one jar to load all jdbc drivers. This PR introduces this change by allowing you to separate each driver path with a comma. This is done like this so it does not break backwards compatibility with previous versions that accepted an string/path. In feature versions the config option should be changed to be an array as more suitable data type for this value. 

Fixes #85

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -96,7 +96,11 @@ module LogStash::PluginMixins::Jdbc
   private
   def load_drivers(drivers)
     drivers.each do |driver|
-      require driver
+      begin
+        require driver
+      rescue => e
+        @logger.error("Failed to load #{driver}", :exception => e)
+      end
     end
   end
 
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -18,10 +18,11 @@ module LogStash::PluginMixins::Jdbc
 
   public
   def setup_jdbc_config
-    # JDBC driver library path to third party driver library.
+    # JDBC driver library path to third party driver library. In case of multiple libraries being
+    # required you can pass them separated by a comma.
     #
     # If not provided, Plugin will look for the driver class in the Logstash Java classpath.
-    config :jdbc_driver_library, :validate => :path
+    config :jdbc_driver_library, :validate => :string
 
     # JDBC driver class to load, for exmaple, "org.apache.derby.jdbc.ClientDriver"
     # NB per https://github.com/logstash-plugins/logstash-input-jdbc/issues/43 if you are using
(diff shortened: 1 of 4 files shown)


purbon:
@talevy would you be so nice to tell me your opinion on this? thanks a lot.

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -92,12 +93,20 @@ def jdbc_connect
     end
   end
 
+  private
+  def load_drivers(drivers)
+    drivers.each do |driver|
+      require driver
```
talevy: should we wrap this with a friendly rescue block for more descriptive error logging if driver jar path is not valid?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -18,10 +18,11 @@ def self.included(base)
 
   public
   def setup_jdbc_config
-    # JDBC driver library path to third party driver library.
+    # JDBC driver library path to third party driver library. In case of multiple libraries being
+    # required you can pass them separated by a comma.
     #
     # If not provided, Plugin will look for the driver class in the Logstash Java classpath.
-    config :jdbc_driver_library, :validate => :path
+    config :jdbc_driver_library, :validate => :string
```
suyograo: @purbon an array in LS also allows a single value. right @jordansissel ?

suyograo:
@purbon Can you update this after review?

lib/logstash/plugin_mixins/jdbc.rb
purbon: for sure! good catch man

lib/logstash/plugin_mixins/jdbc.rb
purbon: It will be, I agree, but then this changes the configuration making all previous ones incompatible. Is this something we want to make?

purbon:
@suyograo updated, only thing to settle is if we make the config option an array or not? if we change it then the change will be backwards incompatible, so I choose a less good, but backwards compatible way. Let me know what do you think.

purbon:
@talevy @suyograo are you ok with this?

talevy:
I think it is fine that we leave it as a `,` separated string. This is how we do a few other configuration options in other plugins. so its not inconsistent.

LGTM

lib/logstash/plugin_mixins/jdbc.rb
jordansissel: @purbon I think changing to :array is still compatible. Here's my thinking:

* `:path` is basically `:string` with an additional validation step to verify the path exists
* `:string` is a single "string"
* `:array` can be a single "string" in the config, or `[ "multiple", "values" ]`
* When an `:array` is given a single "string" like `jdbc_driver_library => "foo"` it is the same as if you had a single-element array: `jdbc_driver_library => [ "foo" ]`

So if we change validate to `:array` as @suyograo proposed, we still are backwards compatible, but make it hopefully more obvious how to specify multiple libraries to load.

Thoughts on using `:validate => :array` ?

PR #87, opened 2015-11-19
Increase log verbosity and levels of information

### LINKED ISSUE #61: Include number of records fetched per statement in debug logging
Currently, at debug logging level, we write out the SQL statements with parameters that are being issued by the plugin, it will also be helpful to write out a count of the # of records returned by each specific statement.

purbon:
As requested in #61 this PR add more information to the debug log statements and also expose the sql_log_level variable from sequel.

Fixes #61

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -73,8 +73,9 @@ module LogStash::PluginMixins::Jdbc
     # documentation page: https://github.com/jeremyevans/sequel/blob/master/doc/opening_databases.rdoc
     config :sequel_opts, :validate => :hash, :default => {}
 
-    # Log level at which to log SQL queries. The default value is info.
-    config :sql_log_level, :validate => :string, :default => "info"
+    # Log level at which to log SQL queries, the accepted values are the common ones fatal, error, warn,
+    # info and debug. The default value is info.
+    config :sql_log_level, :validate => [ "fatal", "error", "warn", "info", "debug" ], :default => "info"
   end
 
   private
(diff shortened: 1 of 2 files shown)


lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -72,6 +72,9 @@ def setup_jdbc_config
     # examples of vendor-specific options can be found in this
     # documentation page: https://github.com/jeremyevans/sequel/blob/master/doc/opening_databases.rdoc
     config :sequel_opts, :validate => :hash, :default => {}
+
+    # Log level at which to log SQL queries. The default value is info.
+    config :sql_log_level, :validate => :string, :default => "info"
```
suyograo: can we restrict this to accepted values? Also can you add this to the doc right above this line?

suyograo: Also @purbon is info the right default? How about warning?

purbon: will add accepted values, and info in the docs. re default, yes it is, because this is a filter that configure sequel (As I understand), if we have warning here as default it will log everything at runtime, we don't want that in normal execution. makes sense?

talevy:
since we are trying to be better at logging and give it first-class importance, should we add tests for this?

purbon:
@talevy yeah, makes sense. will do.

purbon:
@talevy test added, let me know what do you think. Test are added not the best way, but actually I tried the less changing way. We should look for way to really inject dependencies like the logger or also the database connection. This might be food for another improvement here.

talevy:
LGTM

Issue #93, opened 2015-12-07
logstash-input-jdbc to record sql_last_start per query execution


a-k-:
Currently if query is scheduled, sql_last_start is not per run (in case there are errors during query execution).

sql_last_start is set when plugin is registered , updated before query execution and is written to file on stop(). Scheduler is running query execution code that does not flush sql_last_start,not being able to trace the last time query was executed successfully and when there are errors during query execution sql_last_start is still updated. 

Suggesting to tie sql_last_start management with query execution and update sql_last_start after query execution

PR #94, opened 2015-12-07
sql start time per run

a-k-:
Two changes:
 - record sql last start time only on successful query execution. Mostly affected if scheduler is enabled that pulls latest updates based on the last start run. In case if query fails, current updates should be included with the next run
 - flush last start time after each query execution. Handy to trace the last time query has been executed

https://github.com/logstash-plugins/logstash-input-jdbc/issues/93

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -35,8 +35,8 @@ require "yaml" # persistence
 #
 # ==== State
 #
-# The plugin will persist the `sql_last_start` parameter in the form of a
-# metadata file stored in the configured `last_run_metadata_path`. Upon shutting down,
+# The plugin will persist the `sql_last_start` parameter in the form of a 
+# metadata file stored in the configured `last_run_metadata_path`. Upon query execution, 
 # this file will be updated with the current value of `sql_last_start`. Next time
 # the pipeline starts up, this value will be updated by reading from the file. If
 # `clean_run` is set to true, this value will be ignored and `sql_last_start` will be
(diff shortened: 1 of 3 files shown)


suyograo:
@a-k- nice! this is a good tweak. Can you please add a test for this?

a-k-:
great , tests have been added

suyograo:
Tests pass 

```
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}
......................

Finished in 10.65 seconds (files took 1.32 seconds to load)
22 examples, 0 failures

Randomized with seed 54389
```

LGTM

lib/logstash/inputs/jdbc.rb
```diff
@@ -162,22 +162,19 @@ def run(queue)
       @scheduler = Rufus::Scheduler.new(:max_work_threads => 1)
       @scheduler.cron @schedule do
         execute_query(queue)
+        update_state_file()
```
suyograo: minor style issue: `update_state_file` vs `update_state_file()`. We skip parentheses for a method call if the method accepts no arguments.

lib/logstash/inputs/jdbc.rb
```diff
@@ -162,22 +162,19 @@ def run(queue)
...
       end
 
       @scheduler.join
     else
       execute_query(queue)
+      update_state_file()
```
suyograo: We skip parentheses for a method call if the method accepts no arguments.

suyograo:
@a-k- can you address 2 minor style issues and we can merge this.

a-k-:
sure, updated

suyograo:
@a-k- thanks, can you rebase please :)

a-k-:
@suyograo , done

PR #95, opened 2015-12-07
Set application timezone for zoneless datestamps

untergeek:
Fortunately, the Sequel gem already had the necessary functionality.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -1,8 +1,10 @@
 # encoding: utf-8
 # TAKEN FROM WIIBAA
 require "logstash/config/mixin"
+require "time"
+require "date"
 
-# Tentative of abstracting JDBC logic to a mixin 
+# Tentative of abstracting JDBC logic to a mixin
 # for potential reuse in other plugins (input/output)
 module LogStash::PluginMixins::Jdbc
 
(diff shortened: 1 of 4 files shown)


spec/inputs/jdbc_spec.rb
```diff
@@ -202,6 +204,42 @@
...
+
+    after do
+      plugin.stop
+    end
+
+    it "should convert the time to reflect the timezone " do
+      plugin.run(queue)
+      event = queue.pop
+      # This reflects a 6 hour time difference between UTC and America/Chicago
+      expect(event["custom_time"].time).to eq(Time.iso8601("2015-01-01T18:00:00Z"))
```
talevy: I understand this tests that the time are equal, should we also test that the timezone is correct? by formatting it as a string possibly? or fetching the timezone?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -65,6 +67,14 @@ def setup_jdbc_config
     # The amount of seconds to wait to acquire a connection before raising a PoolTimeoutError (default 5)
     config :jdbc_pool_timeout, :validate => :number, :default => 5
 
+    # Timezone conversion.
```
talevy: these docs do a great job describing the behaviors around timezones, but it would be great to be clearer around what setting this parameter will achieve.

spec/inputs/jdbc_spec.rb
```diff
@@ -5,10 +5,13 @@
 require "sequel/adapters/jdbc"
 require "timecop"
 require "stud/temporary"
+require "time"
+require "date"
 
 describe LogStash::Inputs::Jdbc do
+  ENV['TZ'] = "UTC"
```
talevy: Since this affects the global state of the plugin runs and does not get cleaned up between individual spec tests... would it make more sense to use `stub_const` as such in the `:before` block?

```
stub_const('ENV', ENV.to_hash.merge('TZ' => 'UTC'))
```

talevy:
LGTM

PR #97, opened 2015-12-09
Add tzinfo-data and fix tests

untergeek:
The previous attempt to use stub_const did not "unset" between runs.
It is simpler to hard-code a timezone for all tests, to guarantee
that there are no ambiguities.

Because SQL timestamps do not allow for timezone or offset data, this
should not be a limiting factor for the tests.

### DIFF
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -9,6 +9,10 @@ require "time"
 require "date"
 
 describe LogStash::Inputs::Jdbc do
+  # This is a necessary change test-wide to guarantee that no local timezone
+  # is picked up.  It could be arbitrarily set to any timezone, but then the test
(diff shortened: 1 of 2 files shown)


spec/inputs/jdbc_spec.rb
```diff
@@ -9,6 +9,10 @@
 require "date"
 
 describe LogStash::Inputs::Jdbc do
+  # This is a necessary change test-wide to guarantee that no local timezone
+  # is picked up.  It could be arbitrarily set to any timezone, but then the test
+  # would have to compensate differently.  That's why UTC is chosen.
+  ENV["TZ"] = "Etc/UTC"
```
talevy: I would bring back stub_const and place it in the `before :each` block here:

talevy: I overlooked that in the previous PR

talevy:
LGTM

Issue #44, opened 2015-08-25
Upper/Lower Case of Column Names and other Identifiers is not preserved in some Drivers


PhaedrusTheGreek:
In some driver modes, such as MSSQL, the case of column names and other identifiers is not preserved.  This results in output with the incorrect key names.   i.e., ColumnOne is not the same as columnone.

The Sequel library has an API to change the default behavior, but the corresponding option is not available in Logstash configuration.

In my specific case, I was losing capitalization on my MSSQL Column Names.  To workaround, I modified plugin_mixins/jdbc.rb, adding the following line:
```ruby
@database.identifier_output_method = :to_s
```
under the existing Sequel.connect line.

guyboertje:
I worked with Jay @PhaedrusTheGreek on this.  The underlying problem is that Sequel has a downcase converter default that is applied to all field names when processing the resultset.  An option is available on the Sequel class or the database connection to provide a different operation.  The mechanism is <String>.send(:symbol) so we can only use methods available on the String class.

Consider adding a config to give the user a way to control what case conversion is applied to the fields from the resultset, or perhaps the resulting keys in the event.

monster-hkim:
@PhaedrusTheGreek and @guyboertje, do you have the gem file for windows with this workaround in place? I am using the LogStash 2.0 and running into this exact behavior. It would be wonderful if I could download the gem file and use it. Thanks!

monster-hkim:
Ok, figured it out. Works like a charm!

Under the function "prepare_jdbc_connection", add the workaround after

        @database = jdbc_connect()
	add => @database.identifier_output_method = :to_s

gmoskovicz:
@guyboertje are we going to this this, or this is a jdbc mysql specific issue?

Looks like we can have an extra parameter which is `identifier_output_method` and add that as required. Does this makes sense? any other option on how to resolve this ?

gmoskovicz:
According to the [documentation](http://www.rubydoc.info/github/jeremyevans/sequel/Sequel.identifier_output_method%3D) this will happen in many different databases, identifiers will be lowercased, so we should fix this by setting `to_s` or `nil`.

guyboertje:
@gmoskovicz - I was waiting for anyone to pick this up.  I thought even Jay might create a PR. 

Thinking about it though, the reason why Sequel makes it lowercase is to help with the datamapper and activerecord type frameworks that make the identifiers into Ruby method calls.

[...]

PR #100, opened 2015-12-15
Implement force_lowercase_keys, code and tests. See issue 44

guyboertje:
NOTES TO REVIEWERS:
- the Sequel library will downcase all fields by default.
- Some databases allow for mixed case field and table names e.g. MS SQL
- Introduce a new config option `force_lowercase_keys` defaulting to true for backward compatibility.
- In the test env, the Java Derby DB upper cases all identifiers, so this is tested for.

#44

@gmoskovicz @PhaedrusTheGreek - Merry Christmas! :smile:

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -12,13 +12,13 @@ require "yaml" # persistence
 #
 # ==== Drivers
 #
-# This plugin does not come packaged with JDBC driver libraries. The desired 
+# This plugin does not come packaged with JDBC driver libraries. The desired
 # jdbc driver library must be explicitly passed in to the plugin using the
 # `jdbc_driver_library` configuration option.
-# 
+#
 # ==== Scheduling
 #
-# Input from this plugin can be scheduled to run periodically according to a specific 
+# Input from this plugin can be scheduled to run periodically according to a specific
 # schedule. This scheduling syntax is powered by https://github.com/jmettraux/rufus-scheduler[rufus-scheduler].
 # The syntax is cron-like with some extensions specific to Rufus (e.g. timezone support ).
 #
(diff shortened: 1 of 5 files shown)


gmoskovicz:
@guyboertje LGTM

lib/logstash/inputs/jdbc.rb
```diff
@@ -137,6 +137,9 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to save state or not in last_run_metadata_path
   config :record_last_run, :validate => :boolean, :default => true
 
+  # Whether to force the lowercasing of identifier fields
+  config :force_lowercase_keys, :validate => :boolean, :default => true
```
jsvd: according to https://github.com/logstash-plugins/logstash-input-jdbc/issues/44#issuecomment-164436249 shouldn't this be false?

gmoskovicz: @jsvd to provide the same functionality we should set this to true, i know i added the comment but to keep consistent to the old behaviour, the default was to lowercase the keys. We should add this to the documentation, first thing that people will look into when running into this problem is if they are missing a setting.

BTW: should we call this setting force_lowecase_column_names?

spec/inputs/jdbc_spec.rb
```diff
@@ -506,4 +506,39 @@
...
+      db[:ttt].insert(:num => 42, :somestring => "This is a string")
+      plugin.register
+    end
+
+    after do
+      plugin.stop
+      db.drop_table(:ttt)
+    end
+
+    context "when force_lowercase_keys is off (default)" do
```
jsvd: line 527: `    context "when force_lowercase_keys is off (default)" do`

line 535: `    context "when force_lowercase_keys is off" do`

spec/inputs/jdbc_spec.rb
```diff
@@ -506,4 +506,39 @@
...
+      it "the field names are lower case" do
+        plugin.run(events)
+        values = events.first.to_hash.values_at("num", "somestring")
```
jsvd: my suggestion is to make the spec target the names of the keys, so something similar to:

`expect(events.first.to_hash.keys).to eq(["num", "somestring"])`

guyboertje: I'll see your suggestion and raise you a commit :smile:

guyboertje:
What is the vote on @gmoskovicz suggestion for the option name?
`force_lowercase_column_names`

As I am not a user I will go with the consensus. @PhaedrusTheGreek @jsvd

jsvd:
The word force seems unecessary here, just `lowercase_column_names`?

guyboertje:
True dat

PhaedrusTheGreek:
+1 ```lowercase_column_names```.

guyboertje:
Going with `lowercase_column_names`.

suyograo:
LGTM

jsvd:
LGTM, please squash and leave a note in the commit message saying that `force_lowercase_keys => true` was the default behaviour, i.e. `@database.identifier_output_method` is `:downcase` by default

guyboertje:
Sure. Thanks @jsvd, @suyograo, @gmoskovicz and @PhaedrusTheGreek

PR #108, opened 2016-01-05
Allow tracking by column number, not just time

### LINKED ISSUE #57: latest max column_name instead of sql_last_start
At the moment, it is not possible to query database like: "SELECT * FROM TABLE_NAME where id > : last_saved_id"
and not by using built-in parameter "sql_last_start" e.g. "SELECT * FROM TABLE_NAME where timestamp > : sql_last_start".
I don't have timestamp in my table and I want to schedule logstash to have jdbc input and output in elasticsearch.
I dont want to query whole table, I want to query only new entries from database but I don't know how and where to save the "id" from previous input. This id will be used in next scheduled time. 
If you found this worth to be implemented in the future please do.

untergeek:
closes #57

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -161,7 +161,11 @@ module LogStash::PluginMixins::Jdbc
     else
       @database.identifier_output_method = :to_s
     end
-    @sql_last_start = Time.at(0).utc
+    if @use_column_value
+      @sql_last_value = 0
+    else
+      @sql_last_value = Time.at(0).utc
+    end
   end # def prepare_jdbc_connection
 
   public
@@ -175,17 +179,20 @@ module LogStash::PluginMixins::Jdbc
     begin
       parameters = symbolized_params(parameters)
       query = @database[statement, parameters]
-      sql_last_start = Time.now.utc
+      sql_last_value = @use_column_value ? @sql_last_value : Time.now.utc
+      @tracking_column_warning_sent = false
       @logger.debug? and @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters, :count => query.count)
 
       if @jdbc_paging_enabled
         query.each_page(@jdbc_page_size) do |paged_dataset|
           paged_dataset.each do |row|
+            sql_last_value = get_column_value(row) if @use_column_value
             yield extract_values_from(row)
           end
         end
       else
         query.each do |row|
+          sql_last_value = get_column_value(row) if @use_column_value
           yield extract_values_from(row)
         end
       end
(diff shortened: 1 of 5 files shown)


talevy:
we should not remove the tracking of `sql_last_start`. instead, we should allow for keeping track of both. Some queries will probably want to filter on both a specific `tracking_column` and `sql_last_start`.

untergeek:
You think so?  I'm guess that's a possibility, but it seems like a rather uncommon one.

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -175,17 +179,19 @@ def execute_statement(statement, parameters)
...
       parameters = symbolized_params(parameters)
       query = @database[statement, parameters]
-      sql_last_start = Time.now.utc
+      sql_last_value = @use_column_value ? @sql_last_value : Time.now.utc
       @logger.debug? and @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters, :count => query.count)
 
       if @jdbc_paging_enabled
         query.each_page(@jdbc_page_size) do |paged_dataset|
           paged_dataset.each do |row|
+            sql_last_value = row[@tracking_column.to_sym] if @use_column_value
```
talevy: should we log a warning if `@tracking_column` is not present in the result set?

untergeek: That's probably a good idea.

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -175,17 +179,20 @@ def execute_statement(statement, parameters)
     begin
       parameters = symbolized_params(parameters)
       query = @database[statement, parameters]
-      sql_last_start = Time.now.utc
+      sql_last_value = @use_column_value ? @sql_last_value : Time.now.utc
+      @warning_sent = false
```
talevy: since we may introduce additional warnings, maybe this should be named something specific to the column retrieval?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -193,11 +200,26 @@ def execute_statement(statement, parameters)
...
     else
-      @sql_last_start = sql_last_start
+      @sql_last_value = sql_last_value
     end
     return success
   end
 
+  public
+  def get_column_value(row)
+    if !row[@tracking_column.to_sym]
```
talevy: the column may exist, but be `nil`.

I would either warn on `row[@tracking_column.to_sym].nil?` or `row.has_key?(@tracking_column.to_sym)`,
and then the warning message would be unique to either. what do you think?

talevy:
left a few comments, looks good.

leifurhauks:
When using a tracking column, if the statement retrieves multiple rows, which value will be saved? The maximum value? Or the value from the "last" row in the result set?

talevy:
@leifurhauks as per the current implementation[[1](https://github.com/logstash-plugins/logstash-input-jdbc/blob/master/lib/logstash/plugin_mixins/jdbc.rb#L195)], it is the value from the "last" row in the result-set.

PR #128, opened 2016-05-05
add jdbc_password_filepath parameter, with tests and docs

dariko:
add `jdbc_password_filepath` parameter for reading password from an external file.

Must not be used when `jdbc_password` is set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,7 @@ Reading data from MySQL:
 	    jdbc_connection_string => "jdbc:mysql://host:port/database"
 	    jdbc_user => "user"
 	    jdbc_password => "password"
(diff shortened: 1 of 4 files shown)


karmi:
Hi @dariko, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/128.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

dariko:
:) wrong profile:) now they should match

karmi:
@dariko, cool!, works! :)

PR #141, opened 2016-06-17
Increase scheduler sleep frequency

### LINKED ISSUE #134: High CPU usage while waiting for cron schedule
I noticed a high CPU usage - even though it should not be doing anything but waiting for the cron scheduled time.

In the Rufus scheduler there is a :frequency option which by default is 0.3s - could this be increased here? Maybe this would help?

suyograo:
Fixes #134

Rufus: https://github.com/jmettraux/rufus-scheduler#rufusschedulernew-options

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -195,7 +195,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
   def run(queue)
     if @schedule
-      @scheduler = Rufus::Scheduler.new(:max_work_threads => 1)
+      @scheduler = Rufus::Scheduler.new(:max_work_threads => 1,  :frequency => 30)
       @scheduler.cron @schedule do
         execute_query(queue)
         update_state_file


mrjameshamilton:
will this be released?

guyboertje:
@suyograo - After seeing a customer using a 2 second cron schedule, i'm concerned.
We could parse the cron schedule string looking for seconds and not set frequency - wdyt?

jordansissel:
@guyboertje I agree with you.

I'll follow up on the parent ticket. I have ideas for rufus.

jordansissel:
I commented - https://github.com/logstash-plugins/logstash-input-jdbc/issues/134#issuecomment-300517503

At this time, I do not believe we have enough information to decide that rufus's check frequency is the problem, so I would decline this patch.

jordansissel:
I'm closing this for two reasons:

1) The 30 second time will likely make it impossible for users who are currently using cron schedule per-second such as every-2-seconds

2) I am not seeing data pointing that rufus is the reason for high cpu. Let's track #134 and find the origin of the issue.

PR #142, opened 2016-06-21
add more tests for date timezone offsets

talevy:


untergeek:
I already looked this over.  LGTM™

Issue #143, opened 2016-06-22
codec option is not respected in this plugin.


ph:
The plugin doesn't make usage of the `codec` option in the configuration, some database schema are configured with a different charset than UTF-8. When this is the case it could lead logstash to blow up with a `java.nio.charset.UnsupportedCharsetException: ASCII-8BIT` when accessing some values.

This plugin should allow people to use codec to convert their charset to UTF-8.

jordansissel:
-1 on codecs for this plugin. Codecs were intended to help plugins delimit events (like for tcp and stdin inputs) or parse a binary payload (like inputs redis, rabbitmq, etc). In the case of JDBC, the delimiting already happens (query results are in rows), and there's no "binary" payload to parse because we already have structure (columns).

I think this plugin should ensure that any strings created are converted to UTF-8. In practice, the codec does this in some plugins (stdin, etc), but for reasons above codecs aren't a fit for this plugin.

ph:
@jordansissel Agree, but does this mean we will use another option name for it so users can specify the charset they use when converting?

 If we choose to use a new option, we need to make sure we communicate to the user that the `codec` wont be taken into consideration for this specific input. In the current state of LS, codec params is inherited from the base classes and will default to the plain codec.

jordansissel:
Yeah I'm not sure... Maybe we need a way to provide a charset setting per column?

ph:
or have an hybrid solution, if its a string apply the charset to all the columns, if its an hash apply the select charset for the specific column

jordansissel:
hmm, I think that's doable, yeah, though for a configuration we don't yet support 'hash or string' for value validation

purbon:
I think we somehow got the idea to implement an hybrid solution here, right? i'm going to experiment and try to implement it for this bug.

purbon:
Might be we can also leverage sequel (our ORM used there) to make him take care of encoding, see http://sequel.jeremyevans.net/rdoc-plugins/classes/Sequel/Plugins/ForceEncoding.html as example.

purbon:
Aso keep in mind one thing, the way we use the data here, all is expected to be converted in the execute_query method:

```

  def execute_query(queue)
    # update default parameters
    @parameters['sql_last_value'] = @sql_last_value
    execute_statement(@statement, @parameters) do |row|
      event = LogStash::Event.new(row)
      decorate(event)
      queue << event
    end
  end
```

[...]

PR #146, opened 2016-06-30
Add an option to encode string values to a target encoding

purbon:
This PR adds the ability to have uniform encoding for strings read from the database as there are some situations where we might be fetching strings encoded in different formats.

There are situations where the JDBC driver interacting with databases that have globalization features could be fetching data encoded in different formats, for example some in `UTF-8` while others in `ASCII-8BIT`, so prevent this we introduced the usage of `LogStash::Util::Charset.new(@charset)` the same utility used for LS codecs to sort out encoding. With this addition all strings are going to be always in UTF-8 format.

In #143 were discussed some other alternative, for example providing some degree of granularity to let developers chose their encoding. This is a nice idea, but in this PR I propose for now a simpler and more uniform approach for all columns, reducing the config complexity and probable errors, if there is the need we can always introduce a more granular option.

Also keep in mind this is a bug that happen for LS versions that use the ruby event, not the javaone. 

In ruby event:  "String" {ASCII encode} ---> insert into event --> event --> retrieve from event --> "String" {ASCII encode}

In java event:  "String" {ASCII encode} ---> insert into event --> event --> retrieve from event --> "String" {UTF-8 encode}

what do you think?

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -160,6 +160,14 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # The character encoding used in this input. Examples include `UTF-8`
+  # and `cp1252`
+  #
+  # This setting is useful if some of your data is in `Latin-1` (aka `cp1252`)
+  # or in another character set other than `UTF-8`. After the conversion all data
+  # will be in `UTF-8` as the common encoding for LogStash pipeline.
+  config :charset, :validate => ::Encoding.name_list, :default => "UTF-8"
+
   public
 
   def register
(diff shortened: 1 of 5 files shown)


jsvd:
LGTM

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,13 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # The character encoding used in this input. Examples include `UTF-8`
+  # and `cp1252`
+  #
+  # This setting is useful if some of your data is in `Latin-1` (aka `cp1252`)
+  # or in another character set other than `UTF-8`.
+  config :charset, :validate => ::Encoding.name_list, :default => "UTF-8"
```
jsvd: can we make here clear the purpose of this is to convert from whatever we define in :charset to utf8, so that all data is utf8?

colinsurprenant:
replaced by #149

PR #149, opened 2016-07-01
add encoding feature for string retrieved out of the jdbc stament

purbon:
This PR adds the ability to have uniform encoding for strings read from the database as there are some situations where we might be fetching strings encoded in different formats.

There are situations where the JDBC driver interacting with databases that have globalization features could be fetching data encoded in different formats, for example some in UTF-8 while others in ASCII-8BIT, so prevent this we introduced the usage of LogStash::Util::Charset.new(@charset) the same utility used for LS codecs to sort out encoding. With this addition all strings are going to be always in UTF-8 format.

In #143 were discussed some other alternative, for example providing some degree of granularity to let developers chose their encoding. This is a nice idea, but in this PR I propose for now a simpler and more uniform approach for all columns, reducing the config complexity and probable errors, if there is the need we can always introduce a more granular option.

TODO:

* [x] include config to avoid any conversion at all.
* [x] include config option to select only a list of columns to convert.

In resume for this PR, I propose to have:

* An entry point config variable, `enable_encoding` (false by default) that enable the character conversions.
* An array where we can specify a list of columns we want to convert. If some are specified only this will be converted, if empty (default value) all will be converted.

The encoding is selected based on the incoming string encoding to reduce errors when selecting the wrong encoding, or the encoding changes without noticing for the operators. It also helps keeping the config options smaller.

what do you think?

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -261,7 +261,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
   def should_convert(key, value)
     column_keys = @columns_to_encode.keys
-    value.is_a?(String) && (!@default_encoding.nil? || ( @default_encoding.nil? && column_keys.include?(key)) )
+    value.is_a?(String) && (!@default_encoding.nil? || column_keys.include?(key))
   end
 
   def find_converter_for(key, value)
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -160,6 +160,17 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # Enable the usage of encoding conversions, if not enabled no conversion
+  # will be performed. (default => false).
+  # IF enabled string will be converted uniformly to UTF-8 from their source encoding, this
+  # is handy in some situations to avoid problems when non UTF-8 encoded strings are used.
+  config :enable_encoding, :validate => :boolean, :default => false
+
+  # IF enable_encoding is true (default=false) it let's you select which string columns
+  # should be converted to UTF-8. If empty it will convert all columns. To select
+  # a column to be converted, add it's name to this list.
+  config :columns_to_encode, :validate => :array, :default => []
+
   public
 
   def register
@@ -192,7 +203,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
     @jdbc_password = File.read(@jdbc_password_filepath).strip if @jdbc_password_filepath
 
-    @converters = { Encoding::UTF_8 => LogStash::Util::Charset.new("UTF-8") }
+    @converters = { Encoding::UTF_8 => LogStash::Util::Charset.new("UTF-8") } if @enable_encoding
   end # def register
 
   def run(queue)
@@ -222,6 +233,10 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
     # update default parameters
     @parameters['sql_last_value'] = @sql_last_value
     execute_statement(@statement, @parameters) do |row|
+      if @enable_encoding
+        ## do the necessary conversions to string elements
+        row = Hash[row.map { |k, v| [k.to_s, convert(k, v)] }]
+      end
       event = LogStash::Event.new(row)
       decorate(event)
       queue << event
@@ -232,6 +247,25 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
     if @record_last_run
       File.write(@last_run_metadata_path, YAML.dump(@sql_last_value))
     end
+  end
+
+  private
+  # make sure the encoding is uniform over fields
+  def convert(key, value)
+    return value unless should_convert(key, value)
+    converter = find_converter_for value
+    converter.convert(value)
+  end
+
+  def should_convert(key, value)
+    value.is_a?(String) && (@columns_to_encode.empty? || @columns_to_encode.include?(key))
+  end
+
+  def find_converter_for(value)
+    unless @converters.keys.include?(value.encoding)
+      @converters[value.encoding] = LogStash::Util::Charset.new(value.encoding.to_s)
+    end
+    @converters[value.encoding]
   end
 
 end # class LogStash::Inputs::Jdbc
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -166,6 +166,18 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
   # The character encoding for specific columns. This option will override the `:charset` option 
   # for the specified columns.
+  #
+  # Example:
+  # [source,ruby]
+  # ----------------------------------
+  # input {
+  #   jdbc {
+  #     ...
+  #     columns_charset => { "column0" => "ISO-8859-1" }
+  #     ...
+  #   }
+  # }
+  # this will only convert column0 that has ISO-8859-1 as an original encoding.
   config :columns_charset, :validate => :hash, :default => {}
 
   public
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -160,16 +160,12 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
-  # Enable the usage of encoding conversions, if not enabled no conversion
-  # will be performed. (default => false).
-  # IF enabled string will be converted uniformly to UTF-8 from their source encoding, this
-  # is handy in some situations to avoid problems when non UTF-8 encoded strings are used.
-  config :enable_encoding, :validate => :boolean, :default => false
-
-  # IF enable_encoding is true (default=false) it let's you select which string columns
-  # should be converted to UTF-8. If empty it will convert all columns. To select
-  # a column to be converted, add it's name to this list.
-  config :columns_to_encode, :validate => :array, :default => []
+  # If not empty (default={}) it let's you select which string columns
+  # should be converted to UTF-8. To especify a column to be converted you can add the pair
+  # "column_name" => "original_encoding_type" to this hash. If you want to encode your columns
+  # you can use a definition like "_" => "original_encoding_type" and the encoding type will be
+  # applied to all columns.
+  config :columns_to_encode, :validate => :hash, :default => {}
 
   public
 
@@ -203,7 +199,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
     @jdbc_password = File.read(@jdbc_password_filepath).strip if @jdbc_password_filepath
 
-    @converters = { Encoding::UTF_8 => LogStash::Util::Charset.new("UTF-8") } if @enable_encoding
+    @converters = { "UTF-8" => LogStash::Util::Charset.new("UTF-8") } if !@columns_to_encode.empty?
   end # def register
 
   def run(queue)
@@ -233,7 +229,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
     # update default parameters
     @parameters['sql_last_value'] = @sql_last_value
     execute_statement(@statement, @parameters) do |row|
-      if @enable_encoding
+      if enable_encoding
         ## do the necessary conversions to string elements
         row = Hash[row.map { |k, v| [k.to_s, convert(k, v)] }]
       end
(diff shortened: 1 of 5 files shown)


colinsurprenant:
> Also keep in mind this is a bug that happen for LS versions that use the ruby event, not the javaone.

Note that this not correct. The same bug might happen in the Java Event because depending on the source charset, the resulting UTF-8 strings might contain invalidly encoded string.

purbon:
@colinsurprenant Ok, out of my curiousity, can you show me an example? me wanna learn from you all encoding wizards .

purbon:
amended that section from the description due to your comments.

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,17 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
...
+  # will be performed. (default => false).
+  # IF enabled string will be converted uniformly to UTF-8 from their source encoding, this
+  # is handy in some situations to avoid problems when non UTF-8 encoded strings are used.
+  config :enable_encoding, :validate => :boolean, :default => false
+
+  # IF enable_encoding is true (default=false) it let's you select which string columns
+  # should be converted to UTF-8. If empty it will convert all columns. To select
+  # a column to be converted, add it's name to this list.
+  config :columns_to_encode, :validate => :array, :default => []
+
```
colinsurprenant: I don't understand where the source charset is specified? to properly perform transcoding, we need to be able to specify the source charset per column.

lib/logstash/inputs/jdbc.rb
```diff
@@ -232,4 +249,23 @@ def update_state_file
...
+    converter = find_converter_for value
+    converter.convert(value)
+  end
+
+  def should_convert(key, value)
+    value.is_a?(String) && (@columns_to_encode.empty? || @columns_to_encode.include?(key))
+  end
+
+  def find_converter_for(value)
+    unless @converters.keys.include?(value.encoding)
```
colinsurprenant: ah, ok, I see, we rely on the actual String value encoding attribute to create a new converter for this encoding. I am not sure this is a reliable strategy?

lib/logstash/inputs/jdbc.rb
purbon: Could not be retrieved with "foo".encoding ? Could this method be wrong on its answer?

purbon: This is actually how charset to convert is selected here.

lib/logstash/inputs/jdbc.rb
purbon: Yes, this is the strategy. I think should reliable, but wil do some research to probe it here. Makes sense?

purbon:
@colinsurprenant after doing a bit of research I think you're right and we basically can not rely on "string".encoding, so I pushed 5d2acbe09ce9b384bf06df666eec9ce5353a90d4 to make sure users specify the right encoding for their columns. Let me know what do you think

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,13 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # If not empty (default={}) it let's you select which string columns
+  # should be converted to UTF-8. To especify a column to be converted you can add the pair
+  # "column_name" => "original_encoding_type" to this hash. If you want to encode your columns
+  # you can use a definition like "_" => "original_encoding_type" and the encoding type will be
+  # applied to all columns.
```
colinsurprenant: this comment will need to be English-proofed but I will wait until we move forward with the reviews as it may change the configuration options.

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,13 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
...
+  config :columns_to_encode, :validate => :hash, :default => {}
```
colinsurprenant: maybe have a separate `config: default_encoding, :validate  => :string, :default => ""` (or default to UTF8?) or something like that? this way a default charset can be applied and we could have a `config :columns_encoding, :validate => :hash, :default => {}` for specific columns exceptions?

jsvd: +1 on having:
 `default_encoding => nil` that executes no encoding change by default
`column_encoding => {}` for column specific encoding

so the rule would be:
  * if column_encoding exists for column, apply that encoding
  * else apply global_encoding if set. do nothing otherwise

purbon: >  this way a default charset can be applied

then we need also a flag to enable encoding, right? I mean to cover the idea of not encoding at all because no necessary. 

without this flag we might find also problems when encoding from wrong default, right?

purbon: ok, yes we can also use nil as default encoding. will push this changes.

colinsurprenant: no need for a  enable encoding flag, with the defaults of `default_encoding => nil` and `column_encoding => {}` no transcoding should be done at all.

purbon:
@colinsurprenant @jsvd interface request pushed.

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,14 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # Default encoding applied to all columns unless no specific encoding is defined
+  # in the columns_to_encode variable. By default this value is not defined
+  config :default_encoding, :validate => :string
+
+  # It let's you select which string columns should be converted to UTF-8 from a given encoding type.
+  # To especify a column to be converted you can add the pair "column_name" => "original_encoding_type" to this hash.
+  config :columns_to_encode, :validate => :hash, :default => {}
```
colinsurprenant: I suggest `:columns_encoding` instead of  `:columns_to_encode` - it feels more consistent

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,14 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # Default encoding applied to all columns unless no specific encoding is defined
+  # in the columns_to_encode variable. By default this value is not defined
+  config :default_encoding, :validate => :string
```
colinsurprenant: I suggest we keep the same option naming as in the codecs and use `:charset` instead of `default_encoding`.

colinsurprenant: so for the doc comment I suggest something like this:

```
# The character encoding of all columns, leave empty if the columns are already properly UTF-8 
# encoded. Specific columns charsets using :columns_charset can override this setting.
```

lib/logstash/inputs/jdbc.rb
colinsurprenant: so for the doc comment I suggest something like this:

```
# The character encoding for specific columns. This option will override the `:charset` option 
# for the specified columns.
```

lib/logstash/inputs/jdbc.rb
```diff
@@ -232,4 +246,30 @@ def update_state_file
     end
   end
 
+  private
+
+  def enable_encoding
+    !@default_encoding.nil? || (@default_encoding.nil? && !@columns_to_encode.empty?)
```
colinsurprenant: this ca be written as:
```
!@default_encoding.nil? || !@columns_to_encode.empty?
```

lib/logstash/inputs/jdbc.rb
```diff
@@ -232,4 +246,30 @@ def update_state_file
...
+
+  # make sure the encoding is uniform over fields
+  def convert(key, value)
+    return value unless should_convert(key, value)
+    converter = find_converter_for(key, value)
+    converter.convert(value)
+  end
+
+  def should_convert(key, value)
+    column_keys = @columns_to_encode.keys
```
colinsurprenant: extracting the keys at each columns for each row seems inefficient, the `column_keys` can be cached no?

lib/logstash/inputs/jdbc.rb
```diff
@@ -232,4 +246,30 @@ def update_state_file
...
+    value.is_a?(String) && (!@default_encoding.nil? || column_keys.include?(key))
+  end
+
+  def find_converter_for(key, value)
+    encoding = @columns_to_encode[key] || @default_encoding
+    unless @converters.keys.include?(encoding)
+      @converters[encoding] = LogStash::Util::Charset.new(encoding)
+    end
+    @converters[encoding]
+  end
```
colinsurprenant: This method also feels inefficient? Unless I am missing something, the same query should yield the same columns so this `@converters` initialization could be all done on the very first row and then just do a simple lookup in `@converters` for the rest no?

colinsurprenant: Here's a different strategy you could look at: instead of having this `should_convert` / `find_converter_for` dance for all columns and row, just create a conversion strategy for **all** columns with one possible strategy being a "no conversion" or "identity". This conversion strategy for all columns could be initialized at the very first result row, and then simply systematically applied to all other rows and it would be a simple loop like this:

```
transcoded_row = row.inject({}) do |result, (k, v)|
  result[k] = @columns_converters[k].call(v)
  result
end
```

or something like that?

purbon: you're right with https://github.com/logstash-plugins/logstash-input-jdbc/pull/149#discussion_r69488329, I will do the changes.

purbon:
@colinsurprenant applied your request of changes, what do you think now?

lib/logstash/inputs/jdbc.rb
```diff
@@ -160,6 +160,14 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # Whether to force the lowercasing of identifier fields
   config :lowercase_column_names, :validate => :boolean, :default => true
 
+  # The character encoding of all columns, leave empty if the columns are already properly UTF-8 
+  # encoded. Specific columns charsets using :columns_charset can override this setting.
+  config :charset, :validate => :string
+
+  # The character encoding for specific columns. This option will override the `:charset` option 
+  # for the specified columns.
```
purbon: I can do, but here I just c&p your proposal 👍, do you find anything else missing?

spec/inputs/jdbc_spec.rb
```diff
@@ -894,4 +896,82 @@
...
+    before(:each) do
+      allow_any_instance_of(Sequel::JDBC::Derby::Dataset).to receive(:each).and_yield(row)
+      plugin.register
+    end
+
+    after(:each) do
+      plugin.stop
+    end
+
+    it "shuold not convert any column by default" do
```
colinsurprenant: shuold -> should ?

spec/inputs/jdbc_spec.rb
```diff
@@ -894,4 +896,82 @@
...
+
+    context "when all string columns should be encoded" do
+
+      let(:settings) do
+        { "statement" => "SELECT * from test_table",
+          "charset" => "ISO-8859-1" }
+      end
+
+      let(:row) do
+        {"column0" => "foo".force_encoding(Encoding::ISO_8859_1), "column1" => "bar".force_encoding(Encoding::ISO_8859_1), "column2" => 3}
```
colinsurprenant: cosmetic comment: you format this `let` and the previous `let` in 2 different ways - I would pick one formatting. Personally my preference for multi values hash literals initialization is like this:
```ruby
{
  "column0" => "foo".force_encoding(Encoding::ISO_8859_1),
  "column1" => "bar".force_encoding(Encoding::ISO_8859_1),
  "column2" => 3
}
```

spec/inputs/jdbc_spec.rb
```diff
@@ -894,4 +896,82 @@
...
+    context "when only an specific column should be converted" do
+
+      let(:settings) do
+        { "statement" => "SELECT * from test_table",
+          "columns_charset" => { "column1" => "ISO-8859-1" } }
+      end
+
+      let(:row) do
+        {"column0" => "foo", "column1" => "bar".force_encoding(Encoding::ISO_8859_1), "column2" => 3, "column3" => "berlin".force_encoding(Encoding::ASCII_8BIT)}
+      end
```
colinsurprenant: same comment for the let formatiing

lib/logstash/inputs/jdbc.rb
```diff
@@ -232,4 +252,23 @@ def update_state_file
...
+    if @columns_charset[column_name]
+      converter = @converters[@columns_charset[column_name]]
+      converter.convert(value)
+    elsif @charset
+      converter = @converters[@charset]
+      converter.convert(value)
+    else
+      value
+    end
+  end
```
purbon: I agree this code is more compact, but I dislike and thing is error prone to do things like initializations in the conditions like you propose in ``` if (column_name_charset = @columns_charset[column_name]) ```

Can we agree on middle term here and get this in? we can always refactor later on.

purbon: there is a pre initialized converting function created, but in order to select the right converter I have to know the encoding of each column.

lib/logstash/inputs/jdbc.rb
colinsurprenant: ah, never mind, there's an example below :P

lib/logstash/inputs/jdbc.rb
colinsurprenant: well, `if (column_name_charset = @columns_charset[column_name])` might be error prone but if there is no error, there's no error and it's not like an error will magically happen. plus if it is well tested then what can go wrong? 

if you are not comfortable with this style then maybe if you prefer you can break it in 2 lines? again, this is just a simple optimization to avoid useless double hash lookup for the same key.
```ruby
column_name_charset = @columns_charset[column_name]
if column_name_charset
```

purbon:
@jsvd will take over this PR as I won't be able to continue with it next days.

jsvd:
closing this in favor of #151

PR #151, opened 2016-07-06
add encoding feature for string retrieved out of the jdbc stament

jsvd:
replaces https://github.com/logstash-plugins/logstash-input-jdbc/pull/149

addresses concerns of https://github.com/logstash-plugins/logstash-input-jdbc/issues/143

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -261,7 +261,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
   def should_convert(key, value)
     column_keys = @columns_to_encode.keys
-    value.is_a?(String) && (!@default_encoding.nil? || ( @default_encoding.nil? && column_keys.include?(key)) )
+    value.is_a?(String) && (!@default_encoding.nil? || column_keys.include?(key))
   end
 
   def find_converter_for(key, value)
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -273,8 +273,9 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # make sure the encoding is uniform over fields
   def convert(column_name, value)
     return value unless value.is_a?(String)
-    if @columns_charset[column_name]
-      converter = @converters[@columns_charset[column_name]]
+    column_charset = @columns_charset[column_name]
+    if column_charset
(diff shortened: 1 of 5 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -232,4 +264,24 @@ def update_state_file
     end
   end
 
+  private
+
+  def enable_encoding?
+    !@charset.nil? || !@columns_charset.empty?
```
colinsurprenant: small perf improvement suggestion: since this method is called on every row, we could memoize the `!@charset.nil? || !@columns_charset.empty?` expression since it will never change

spec/inputs/jdbc_spec.rb
```diff
@@ -894,4 +896,97 @@
...
   end
+
+  context "when encoding of some columns need to be changed" do
+
+    let(:settings) {{ "statement" => "SELECT * from test_table" }}
+    let(:events)   { [] }
+    let(:row) do
+      {"column0" => "foo",
+       "column1" => "bar".force_encoding(Encoding::ISO_8859_1),
+       "column2" => 3}
```
colinsurprenant: this is cosmetic nitpicking but I see you "standardized" the literal hash initialization format for the last spec context but not for the first ones. it is all valid but it is nicer to adopt just one style...

lib/logstash/inputs/jdbc.rb
```diff
@@ -173,6 +193,8 @@ def register
       end
     end
 
+    @enable_encoding = !@charset.nil? || !@columns_charset.empty?
```
colinsurprenant: 👍

lib/logstash/inputs/jdbc.rb
```diff
@@ -220,6 +250,10 @@ def execute_query(queue)
     # update default parameters
     @parameters['sql_last_value'] = @sql_last_value
     execute_statement(@statement, @parameters) do |row|
+      if enable_encoding?
```
colinsurprenant: yeah. we could skip the method call and just do `if @enable_encoding` here... but fine if you prefer it like this.

jsvd: jruby should be smart enough to inline this eventually. anyway I see your point but I'll leave as is for now.

colinsurprenant:
LGTM - did not run tests

jsvd:
Aaaand I should have squashed this..... 👎

PR #152, opened 2016-07-07
Add options in the config to control how the character encoding will work.

jsvd:
Addresses concerns of #143 (backport of #151)


suyograo:
LGTM

PR #155, opened 2016-08-01
Add metrics to the jdbc input operations

purbon:
Add metrics to track:
- # events processed.
- # connections open.
- # operations performed with the database
- # errors that happened during execution.
- small others last sql_value, query size, etc (useful to debug behaviour).

related to https://github.com/elastic/logstash/issues/5607 https://github.com/elastic/logstash/issues/5717

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -183,6 +183,8 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   public
 
   def register
+    @metric_errors = metric.namespace(:errors)
+
     require "rufus/scheduler"
     prepare_jdbc_connection
 
@@ -247,6 +249,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   private
 
   def execute_query(queue)
+    metric.increment(:queries)
     # update default parameters
     @parameters['sql_last_value'] = @sql_last_value
     execute_statement(@statement, @parameters) do |row|
@@ -257,11 +260,13 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
       event = LogStash::Event.new(row)
       decorate(event)
       queue << event
+      metric.increment(:events)
     end
   end
 
   def update_state_file
     if @record_last_run
+      metric.increment(:state_file_updates)
       File.write(@last_run_metadata_path, YAML.dump(@sql_last_value))
     end
   end
@@ -279,9 +284,11 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
     if column_charset
       converter = @converters[column_charset]
       converter.convert(value)
+      metric.increment(:encoding_conversions)
     elsif @charset
       converter = @converters[@charset]
       converter.convert(value)
+      metric.increment(:encoding_conversions)
     else
       value
     end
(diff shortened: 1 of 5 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -279,9 +284,11 @@ def convert(column_name, value)
     if column_charset
       converter = @converters[column_charset]
       converter.convert(value)
+      metric.increment(:encoding_conversions)
     elsif @charset
       converter = @converters[@charset]
       converter.convert(value)
+      metric.increment(:encoding_conversions)
```
ph: I am not sure about the usefulness of theses metrics?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -159,6 +161,8 @@ def prepare_jdbc_connection
       raise LogStash::ConfigurationError, "#{e}. #{message}"
     end
     @database = jdbc_connect()
+    metric.increment(:connections)
+
```
ph: This seems to be only called at register time? so it will be always 1 ?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -220,6 +227,7 @@ def execute_statement(statement, parameters)
       success = true
     rescue Sequel::DatabaseConnectionError, Sequel::DatabaseError => e
       @logger.warn("Exception when executing JDBC query", :exception => e)
+      @metric_errors.increment(:jdbc_query_errors)
```
ph: I think it would be better if we remove the mention of jdbc and maybe have
`DatabaseConnectionError` and `DatabaseError` in two different metric?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -115,14 +115,16 @@ def jdbc_connect
...
+          @logger.error("Failed to connect to database. #{@jdbc_pool_timeout} second timeout exceeded. Trying again.")
+          @metric_errors.increment(:connection_retries)
         end
       rescue Sequel::Error => e
         if retry_attempts <= 0
           @logger.error("Unable to connect to database. Tried #{@connection_retry_attempts} times", :error_message => e.message, )
           raise e
         else
           @logger.error("Unable to connect to database. Trying again", :error_message => e.message)
+          @metric_errors.increment(:connection_retries)
```
jordansissel: The `retries` name here feels like not the right one. Maybe we call this just `:connection` ? So it may read as "connection errors" instead of "connection retry errors".

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -201,8 +205,11 @@ def execute_statement(statement, parameters)
       parameters = symbolized_params(parameters)
       query = @database[statement, parameters]
       sql_last_value = @use_column_value ? @sql_last_value : Time.now.utc
+      metric.gauge(:sql_last_value, sql_last_value)
+
       @tracking_column_warning_sent = false
       @logger.debug? and @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters, :count => query.count)
+      metric.gauge(:rows_in, query.count)
```
jordansissel: I propose just `rows` (this is an input, so the `in` feels implied to me)

Issue #125, opened 2016-04-17
tracking_column with timestamp


nkelly75:
I would like to use a timestamp column as my tracking_column however it seems like use_column_value only works for numeric columns. My config includes:

```
    use_column_value => true
    tracking_column => "updt"
```
My statement includes:
`  AND updt >= :sql_last_value
`

For a fresh start I remove ~/.logstash_jdbc_last_run. Because use_column_value=true, the plugin writes 0 to ~/.logstash_jdbc_last_run. The first attempt to query the DB fails because the query uses zero instead of 1970..

```
{:timestamp=>"2016-04-17T10:25:00.163000-0400", :message=>"Exception when executing JDBC query", :exception=>#<Sequel::DatabaseError: Java::JavaSql::SQLSyntaxErrorException: ORA-00932: inconsistent datatypes: expected DATE got NUMBER
>, :level=>:warn}
```

I was able to work around this by temporarily tweaking line 183 of plugin_mixins/jdbc.rb:

```
    if @use_column_value
      @sql_last_value = Time.at(0).utc
    else
      @sql_last_value = Time.at(0).utc
    end

```

The plugin would then start with 1970 and subsequently start using a value retrieved from the DB. However the tweak above would clearly break things for users who really want numeric column values.

Perhaps one more config setting is required e.g. column_value_timestamp which if true would use 1970 epoch date but if false would behave as today i.e. 0.

nkelly75:
Note: originally I was not using tracking_column instead I was comparing my 'updt' column to sql_last_value i.e. based on last scheduled run of logstash. However this lead to 'gaps' in data.

The DB acting as my source is updated periodically ~every 15 mins (it is downstream from the system of record). I had LogStash configured on a a 5 min interval. Logstash would lose roughly two thirds of data in this scenario - if it found nothing new in the source it would still update sql_last_value metadata..

PR #156, opened 2016-08-01
Add bool config for tracking column as timestamp

growse:
Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: https://www.elastic.co/contributor-agreement/

This allows plugin users to set `use_column_value` on tracking columns
where the column type is a timestamp. Currently, this is only possible
using a numeric column, but there are use cases where the user wants to
select rows from a table where the column value is greater than the last
seen timestamp value of that column.

It turns out that this behaviour works fine with simply setting
`use_column_value` to true on the required column. The only change
needed is to add a flag indicating that on a clean run, the default min
value should be set to `Time.at(0).utc` rather than `0`.

This should close logstash-plugins/logstash-input-jdbc#125

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -151,8 +151,8 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
   # If tracking column value rather than timestamp, the column whose value is to be tracked
   config :tracking_column, :validate => :string
 
-  # Tracking column is a timestamp value to be tracked, rather than an int
-  config :tracking_column_is_timestamp, :validate => :boolean, :default => false
+  # Type of tracking column. Currently only "numeric" and "timestamp"
+  config :tracking_column_type, :validate => ['numeric', 'timestamp'], :default => 'numeric'
 
   # Whether the previous run state should be preserved
   config :clean_run, :validate => :boolean, :default => false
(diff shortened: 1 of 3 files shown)


talevy:
@growse this is great! thanks for the enhancement!

what are your thoughts on a `tracking_column_type` config option that currently only has support for `numeric` and `timestamp`? That way we can also be more explicit about the fact that it will be defaulted to a `0` for numeric types. what do you think?

growse:
Sounds good - I renamed the config value to `tracking_column_type` with current allowable `numeric` and `timestamp` which should allow for some future flexibility.

growse:
Hmm. Looks like the tests sometimes work, and sometimes don't - I didn't change anything about the test other than the config. I'm not a ruby expert, so any help in fixing the failing test would be appreciated...

talevy:
@growse I'll take a look at the updates.

regarding the tests, are they failing locally often? We tend to rely on specific timing properties of the system that it is running on for them to pass. Looks like things are passing on our continuous integration server as noted by the travis-ci check!

growse:
It's just the test ordering that's the problem. Given that rspec randomizes the order each time, it meant that sometimes the suite passed, and sometimes it failed. The latest commit should get around that by being a bit more strict about the type of `sql_last_value` when the tracking column type is set to "timestamp"

talevy:
the test should clear the database before each test instance though

talevy:
LGTM. thanks for the contribution. I will clear up any of these state problems in the tests before releasing

matarrese:
Hi all, Is it possible to have a new release that includes this change merged?

PR #159, opened 2016-08-26
fix to be compatible with LS 5.0 core

talevy:
LS 5.0 introduces log4j2 logging, so this PR updates the plugin to support that.


ph:
LGTM, travis passes.

PR #170, opened 2016-10-18
Update elasticsearch output plugin name

gaving:
As per https://www.elastic.co/guide/en/logstash/current/plugins-outputs-elasticsearch.html#_synopsis_57


gaving:
Nevermind, I see 1.5 is picked by the branch flag and these are for 2.0

"There's other breaking changes in the config for 2.0, notably "host" becoming "hosts" and taking an array."

Closing.

PR #175, opened 2016-10-30
Added section about using multiple SQL statements

hskoro:
As discussed in https://discuss.elastic.co/t/sql-last-value-doesnt-reiterate-with-multiple-jdbc-inputs/63523, this PR represents a doc update to include some guidance on using multiple SQL statements and the importance of last_run_metadata_path parameter when multiple statements also use the sql_last_value parameter for storing state.


suyograo:
LGTM

PR #176, opened 2016-11-10
Introduce statements_directory parameter

boernd:
We are using this plugin mainly to gather database metrics (session statistics, tablespace usage, business data inconsistencies, etc.). When you have a lot of sql files/statements then for each input a database connection is opened which does not scale very well. The `statements_directory` parameter would allow to place several sql files within a directory which get executed together (e.g. in combination with the schedule option) only using one connection.


mrjameshamilton:
Something like this is needed badly. We have ~500 SQL connections in our config, so ~500 database connections. The problem is also needing to paramerise the schema name, as the queries are the same but on different schemas.

Issue #103, opened 2015-12-18
jdbc_fetch_size no effect?


LxiaoGirl:
logstash 1.5.5
jdbc 1.0.0

error log:
java.lang.OutOfMemoryError: Java heap space
Dumping heap to java_pid24312.hprof ...
Heap dump file created [1159466364 bytes in 21.477 secs]

i read 
https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html#plugins-inputs-jdbc-jdbc_page_size
add parameter jdbc_fetch_size.
but,no effect.
I want to increase memory?

jaykay153:
This isn't working for me either using postgresql. Looking at the postgres documentation, the code needs to explicitly set the connection autocommit mode to false e.g. conn.setAutoCommit(false);
https://jdbc.postgresql.org/documentation/91/query.html#fetchsize-example

For data size of 10 million plus, this is a must (jdbc_paging_enabled is too slow).

jaykay153:
Can this be prioritized please?

suyograo:
thanks for the issue, we'll get this soon.

LxiaoGirl:
fixed?

edeak:
I guess pagination could be made faster if the following statement is executed:
select t.* from (select id from t where id > :LAST_ID limit :LIMIT) q join t on q.id = t.id

I use this trick to iterate through millions of records and its quite fast, the only thing to be solved is getting the last id and use it as a variable inside the plugin.

untergeek:
> the only thing to be solved is getting the last id and use it as a variable inside the plugin.

This was committed in #108 and can be gained by updating to version 3.0.0 of the plugin.

Some interim documentation: https://github.com/logstash-plugins/logstash-input-jdbc/issues/57#issuecomment-169436652

jaykay153:
Many tables don't have the last ID column. What about the jdbc_fetch_size fix please?

jjunqueira:
+1, please fix! 

The paging is slower and because of the way that the count query is implemented for paging(just by wrapping the defined statement) I am seeing Incorrect Key File errors in MySQL.

An alternative might also be allowing a way to define the count query statement?

Let me know if I can do anything to help. Thank you for your hard work on this project!

rspuler:
Serious problem for me too. This plugin does not work with postgresql driver and full import of big tables. Allways results in an OutOfMemoryError due to the reason that  jaykay153 has given.

[...]

PR #200, opened 2017-04-07
Fix jdbc_fetch_size usage with postgresql (#103)

izhigalko:
According to documentation (https://jdbc.postgresql.org/documentation/head/query.html) we need to autocommit off if setFetchSize is used. So i wrap statement execution in transaction with Sequel::JDBC::Transaction cls and rollback transaction after all.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -143,6 +143,7 @@ module LogStash::PluginMixins::Jdbc
     require "java"
     require "sequel"
     require "sequel/adapters/jdbc"
+    require "sequel/adapters/jdbc/transactions"
     load_drivers(@jdbc_driver_library.split(",")) if @jdbc_driver_library
 
     begin
@@ -157,6 +158,7 @@ module LogStash::PluginMixins::Jdbc
       raise LogStash::ConfigurationError, "#{e}. #{message}"
     end
     @database = jdbc_connect()
+    @database.extend(Sequel::JDBC::Transactions)
     @database.extension(:pagination)
     if @jdbc_default_timezone
       @database.extension(:named_timezones)
@@ -207,9 +209,19 @@ module LogStash::PluginMixins::Jdbc
       @tracking_column_warning_sent = false
       @logger.debug? and @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters)
 
-      if @jdbc_paging_enabled
-        query.each_page(@jdbc_page_size) do |paged_dataset|
-          paged_dataset.each do |row|
+      @database.transaction do
+        if @jdbc_paging_enabled
+          query.each_page(@jdbc_page_size) do |paged_dataset|
+            paged_dataset.each do |row|
+              sql_last_value = get_column_value(row) if @use_column_value
+              if @tracking_column_type=="timestamp" and @use_column_value and sql_last_value.is_a?(DateTime)
+                sql_last_value=Time.parse(sql_last_value.to_s) # Coerce the timestamp to a `Time`
+              end
+              yield extract_values_from(row)
+            end
+          end
+        else
+          query.each do |row|
             sql_last_value = get_column_value(row) if @use_column_value
             if @tracking_column_type=="timestamp" and @use_column_value and sql_last_value.is_a?(DateTime)
               sql_last_value=Time.parse(sql_last_value.to_s) # Coerce the timestamp to a `Time`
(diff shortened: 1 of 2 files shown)


karmi:
Hi @izhigalko, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/200.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

izhigalko:
> Hi @izhigalko, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git commit. Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

Thanks. I added and verified second email to my profile.

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -205,26 +207,29 @@ def execute_statement(statement, parameters)
       query = @database[statement, parameters]
       sql_last_value = @use_column_value ? @sql_last_value : Time.now.utc
       @tracking_column_warning_sent = false
-      @logger.debug? and @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters, :count => query.count)
```
andrewvc: Why was `:count` removed? Did it execute an extra `COUNT` or was there some other reason?

lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -205,26 +207,29 @@ def execute_statement(statement, parameters)
...
+      @logger.debug? and @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters)
 
-      if @jdbc_paging_enabled
-        query.each_page(@jdbc_page_size) do |paged_dataset|
-          paged_dataset.each do |row|
+      @database.transaction do
```
andrewvc: Thanks for describing this change in the PR. We really need this documented here as well.

Do you know how this will play with other databases? I'm not anticipating problems ATM, but I'm curious what your thoughts are. I'm wondering if this codepath should be PG only.

andrewvc (changes requested):
This is an awesome PR. I've listed some possible improvements.

lib/logstash/plugin_mixins/jdbc.rb
izhigalko: Count removed cause it slow down huge queries execution when debug enabled. My query have more than 5 subqueries in fields and i wait count execution about 5 hours before remove it.

lib/logstash/plugin_mixins/jdbc.rb
izhigalko: I think its ok with other dbs cause used standard JDBC transaction mechanic with Sequel::JDBC::Transactions, and all dbs that i knew uses transaction mechanic. I can rewrite code for use only with PG driver if u wish.

andrewvc: @izhigalko no, I think that makes sense. I do think it needs a comment to explain why we wrap with a transaction then rollback for seemingly no reason however. If you put that comment in we can move forward :)

andrewvc:
@izhigalko if you can just add in that comment I think we can move forward :)

andrewvc (approved):
LGTM

andrewvc (changes requested):
@izhigalko it looks like the build is failing: https://travis-ci.org/logstash-plugins/logstash-input-jdbc/builds/228698596?utm_source=github_status&utm_medium=notification Can you take a look and fix? We'll need a passing build to merge :)

symbyte:
I'm very interested in seeing this move forward. I cloned locally to see if I could help fix the failing test, but when I ran them they all passed. Perhaps there was an issue with the `Mock Connection` module in the `sequel` dependency that has since been resolved? It seemed really weird anyway that the the tests failed at the addition of a commend. 

Any chance we could run the build on this branch again to see if the tests pass?

izhigalko:
@andrewvc ok, fix this. Test failed cause it uses mock database driver which not implements all needed methods

andrewvc (approved):

phillycheeze:
Possible to have this revisited? The jdbc_fetch_size still doesn't work for postgresql.

caseydm:
Hi all is jdbc_fetch_size working with postgresql now? I just using it in a project after switching from redshift and am running into memory errors. It worked great in redshift.

PR #185, opened 2017-01-13
auto reconnect

tac0x2a:
I implemented auto reconnect to database in running. 

If disconnected to database, it will attempt close and re-open connection.

PR #204, opened 2017-04-28
Cleanup test issues

andrewvc:
This fixes some test failures introduced when merging #185 . These failures weren't the fault of the PR, but rather due to a change in LS core where 'tags' was being incorrectly included in the `to_hash` method. Now that it's fixed in core the tests in this plugin were failing again.

This also explicitly imports the Sequel mangling extension. Failing to do so results in deprecation warnings. This feature is not deprecated, but its automatic importation is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 4.2.0
+ - Automatically reconnect on connection issues
+ - Fix test failures
+ - Explicitly load identifier mangling from Sequel to prevent
+   warning logs
+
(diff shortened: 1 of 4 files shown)


colinsurprenant:
I don't see where the change for "Automatically reconnect on connection issues" is?

andrewvc:
@colinsurprenant that was merged in #185 but is not yet released. I included it in the changelog for that reason :)

colinsurprenant:
LGTM

PR #217, opened 2017-06-07
Fix broken jdbc_pass_word_filepath

yangineer:
This fixes the type mismatch error on `@jdbc_password` caused by the` jdbc_password_filepath` feature.
`@jdbc_password` should be of type `LogStash::Util::Password` instead of `string`.

Now the jdbc_password_filepath feature should work as intended.

### DIFF
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -232,7 +232,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
       raise(LogStash::ConfigurationError, "Only one of :jdbc_password, :jdbc_password_filepath may be set at a time.")
     end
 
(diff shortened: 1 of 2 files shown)


karmi:
Hi @yangineer, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/217.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

karmi:
Hi @yangineer, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/217.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

ehvattum:
Please make a new release with this fix.

Issue #220, opened 2017-06-20
Two bugs in "#open_jdbc_connection" when rescuing a DatabaseConnectionError


guyboertje:
In the `open_jdbc_connection` method.

1. `close_jdbc_connection` is called on `@database` - it has no such method.
2. The logged warning does not capture valuable JDBC Driver messages/trace that may be in the DatabaseConnectionError instance.

avihaleva:
Does reconnect works in any scenario ? Looks like when their is any error on the connection, the connection will never recover. Is there any workaround for this issue ?  Can I cause Logstash to terminate on such error ?

franklauterwald:
@avihaleva I just created pull request #231 which solves the reconnection issue for us. Can you try if it helps you? It is supposed to fix reconnecting

guyboertje:
Closed with https://github.com/logstash-plugins/logstash-input-jdbc/pull/231

guyboertje:
@avihaleva @franklauterwald \
New version of plugin published, thank you.

PR #231, opened 2017-08-18
Mark database connection as dead even when disconnecting fails

franklauterwald:
The jdbc input pluging recently got support for automatically reconnecting when a database connection is lost. In this case it tries to cleanly close the old connection by calling disconnect, then setting its @database reference to nil.
@database being nil is then used as an indication that a new connection needs to be opened.
Calling disconnect leads to a call to java.sql.Connection.close() which may throw an SQLException.

In that case @database remains a reference to the broken connection and we try to reuse that.

This PR adds an exception handler to ensure that the database connection is marked dead even when disconnect throws an exception.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.2.4
+  - [#220](https://github.com/logstash-plugins/logstash-input-jdbc/issues/220) Log exception when database connection test fails
+  - Database reconnect: Mark old connection as dead even when clean disconnect fails.
+
 ## 4.2.3
   - Fix some documentation issues
 
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -170,7 +170,7 @@ module LogStash::PluginMixins::Jdbc
     begin
       @database.test_connection
     rescue Sequel::DatabaseConnectionError => e
-      @logger.warn("Failed test_connection.")
+      @logger.warn("Failed test_connection.", :exception => e)
       close_jdbc_connection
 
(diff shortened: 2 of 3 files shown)


guyboertje:
I do like your patch.

In order to properly close #220, please can you add the exception to the logging statement on line 173 of the same file?
`@logger.warn("Failed test_connection.", :exception => e)`

franklauterwald:
Thank you @guyboertje. I will do so tomorrow. After that, I´ll be on vacation until Sep 5.

Btw.: the elastic contributor license agreement page was broken for me earlier. It seems to work now, so I´ll take care about that too (although the patch is so trivial that copyright does not apply anyway)

guyboertje:
What email did you sign the Agreement with? I think it needs to be the one attached to your Github account.
You should also bump the patch part of the version number in the `.gemspec` file and add a section to the `CHANGELOG.md` file.

franklauterwald:
1) no address at all. After choosing "individual contributor" I was forwarded to an empty page showing just the generic elastic links. Maybe just a browser hickup.
2) ok, will do as you say. Yay, finally learning ruby dev practices 101 ;-)

karmi:
Hi @franklauterwald, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/231.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

franklauterwald:
Hi @karmi, thanks for the reminder. Embarrassingly, one of my machines still had the mail address of a former employer.
I have amended the commits and force-pushed, so I hope all is well now.

Issue #166, opened 2016-09-17
Not optimal sql query when using jdbc_page_size


pdorofiejczyk:
I have prepared following logstash config

```
input {
 jdbc {
   jdbc_driver_library => "mysql-connector-java-5.1.39-bin.jar"
   jdbc_driver_class => "com.mysql.jdbc.Driver"
   jdbc_connection_string => "jdbc:mysql://dbhost:3306/mydb"
   jdbc_user => "user"
   jdbc_password => "password"
   statement => "SELECT * FROM test"
   jdbc_paging_enabled => "true"
   jdbc_page_size => "10000"
 }
}
output {
  stdout {  codec => json_lines  }
}
```

When I run logstash and check `SHOW PROCESSLIST;` on my db, I see following sql statements:

```
SELECT * FROM (SELECT * FROM test) as j LIMIT 10000 OFFSET 10000
```

This query reads all rows from table on each page (a few milions rows in my case) so this paging strategy is not optimal. It should do queries presented below.

```
SELECT * FROM test LIMIT 10000 OFFSET 10000
```

axhiao:
And it is too slow! Is there a proper way to perform sql "select * from tbl" when the amount of table is large?

catinred2:
is there any workaround for this?

axhiao:
There is a solution for this:
jdbc_connection_string => "jdbc:mysql://localhost:3306/mydb?useCursorFetch=true"
jdbc_fetch_size => 50000

pdorofiejczyk:
useCursorFetch will use server-side cursors, which in MySQL are implemented using temporary tables

jbboehr:
In MySQL, a table like this:

```sql
CREATE TABLE `videos` (
  `video_id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `ts` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  PRIMARY KEY (`video_id`),
  KEY `ts` (`ts`)
) ENGINE=InnoDB
```

the optimal queries should be, for count and fetch respectively:

```sql
SELECT COUNT(ts) FROM videos WHERE ts >= :sql_last_value LIMIT 1
```
```sql
SELECT * FROM videos WHERE ts >= :sql_last_value ORDER BY ts ASC LIMIT 1000
```

[...]

PR #232, opened 2017-08-31
Manually take care of the pagination instead of using Sequel

chaodhib:
Fixes https://github.com/logstash-plugins/logstash-input-jdbc/issues/166

As explained [here](https://github.com/logstash-plugins/logstash-input-jdbc/issues/166#issuecomment-325975178), Sequel does not parse any native SQL given to it. Therefore, it does not add properly the LIMIT and OFFSET clauses. The only solution is to do it manually. It's the approach I took.

Apologies in advance for the code. I don't usually program in Ruby.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -219,23 +219,32 @@ module LogStash::PluginMixins::Jdbc
     begin
       parameters = symbolized_params(parameters)
       open_jdbc_connection if @database == nil
-      query = @database[statement, parameters]
       sql_last_value = @use_column_value ? @sql_last_value : Time.now.utc
       @tracking_column_warning_sent = false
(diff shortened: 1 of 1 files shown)


karmi:
Hi @chaodhib, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/232.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

chaodhib:
@Hi. My apoligies. I will take care of that shortly.

chaodhib:
Done.

chaodhib:
I'm not really sure why I have a failing test. It may be because the LIMIT or the OFFSET clause does not work on every DB and to be more specific on the database used for the test?

guyboertje:
Yep, screwed.

We chose to specify a SQL statement as it is the most flexible arrangement.

PR #243, opened 2017-10-25
Travis - add 6.0 build and remove default JRuby 1.7 build

jakelandis:
This change was motivated by https://github.com/elastic/logstash/issues/8518 since when run without the LOGSTASH_BRANCH  environment specified, it pulls the latest released dev-utils, which in turn pulls the 5.5.1-snapshot of logstash core which is now incompatible with 1.3.5 of dev-utils. 

Rather then untangling the knot that is devutils.... this change side steps the issues by simply deferring the supported pipeline build dependency resolution which should always work, else the main build is broken.

We are loosing testing against a 'release' version of logstash, but in actuality, we are only loosing testing against 5.5.1-snapshot due to the incorrect dependency convergence of dev-utils.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,15 +2,14 @@
 sudo: false
 language: ruby
 cache: bundler
-env: 
-rvm:
-- jruby-1.7.25
 matrix:
   include:
-  - rvm: jruby-9.1.10.0
+  - rvm: jruby-9.1.13.0
     env: LOGSTASH_BRANCH=master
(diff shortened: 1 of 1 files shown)


jordansissel:
LGTM - Travis is green for all 4 branches.

jakelandis:
An example of a failure that motivated this change:

https://travis-ci.org/logstash-plugins/logstash-input-jdbc/jobs/292245059

PR #252, opened 2017-12-06
Fix connection and memory leak on pipeline reload.

### LINKED ISSUE #251: Connection and memory leak when pipeline is reloaded
Testing against `5.6.4` with `4.3.0` of this plugin. 
 
Everytime a pipeline is reloaded a connection is leaked. 

For example a healthy non-reloaded pipeline doing jdbc work:
image

Then reload the pipeline ~10 times and the open/closed connections gap widens, and the number of noncollectable objects increase too. 
image

Since we only use 1 connection (by default) and as of `4.3.0` open/close connections per statement, the gap between open and close should always be within 1. 

The root cause here is that the underlying library does not guarantee a connection in use will actually get disconnected.  From the underlying library disconnect method: "This method has the effect of disconnecting from the database, assuming that no connections are currently being used. "

We do not offer any assurances that the pipeline thread during a reload will not try to call disconnect while the connection is use. If that happens the library attempts to disconnect and looses it reference to the connection (thinking it is closed), but when in fact it is still open (since it is in use). Hence the connection leak. All the objects associated with the connection are leaked, hence the memory leak. 

This issue will only impact consumers that reload frequently since the connection and memory leak is directly related to the reload frequency. 

Note - `4.3.0` did not introduce this issue.

jakelandis:
When the pipeline reload thread closes the connection, but the connection is still active (becuase is executing a query), the connection and associated memory will leak. The underlying library does not make any assurances that the connection will indeed be closed if currently in use, and observation shows this to be true. The fix here is to ensure that the a thread can not attempt close the connection (it will block) while the connection is in use. In this case that thread is the pipeline thread and the reload will now block until the current query finishes (or errors).

Fixes #251

Youkit evidence of connection behaving correctly.

A few minutes in with a couple reloads:
image

~30 minutes with ~30 reloads
image

The uncollected memory still grew some...but not nearly as much as it did prior.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -4,6 +4,8 @@ require "logstash/config/mixin"
 require "time"
 require "date"
 
+java_import java.util.concurrent.locks.ReentrantLock
+
 # Tentative of abstracting JDBC logic to a mixin
 # for potential reuse in other plugins (input/output)
 module LogStash::PluginMixins::Jdbc
@@ -191,6 +193,7 @@ module LogStash::PluginMixins::Jdbc
 
   public
   def prepare_jdbc_connection
+    @connection_lock = ReentrantLock.new
     if @use_column_value
       case @tracking_column_type
         when "numeric"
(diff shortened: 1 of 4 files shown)


lib/logstash/plugin_mixins/jdbc.rb
```diff
@@ -206,50 +209,58 @@ def prepare_jdbc_connection
...
     rescue => e
       @logger.warn("Failed to close connection", :exception => e)
+    ensure
+      @connection_lock.unlock
     end
   end
 
   public
   def execute_statement(statement, parameters)
-    success = false
```
jakelandis: github diff sucks....here is the real change:
image

jakelandis:
@yaauie - thanks (TIL about `?w=1`)

PR #256, opened 2017-12-14
Fix thread and memory leak on pipeline reload.

### LINKED ISSUE #255: Thread leak (and likely memory leak) on Logstash reload
When Logstash pipelines are reloaded via the a configuration change, the [rufus](https://github.com/jmettraux/rufus-scheduler) jobs thread is not shut down. 

Profiling via YourKit (filtering live threads by rufus):
After 1 reload:
image

After another reload:
image

..and another reload:
image

The root cause here is that the rufus stop/shutdown is defined as:
```
    def shutdown(opt=nil)

      @started_at = nil

      #jobs.each { |j| j.unschedule }
        # provokes https://github.com/jmettraux/rufus-scheduler/issue/98
      @jobs.array.each { |j| j.unschedule }

      @work_queue.clear

      if opt == :wait
        join_all_work_threads
      elsif opt == :kill
        kill_all_work_threads
      end

      unlock
    end
```
...but we are not sending any options to the shutdown method, so the worker threads never get killed. Any objects associated with that thread would considered to be a memory leak. ..._pure subjecture is that JDBC drivers may use thread locals to hold cached state which could lead to a decent memory leak over time and volume_

jakelandis:
When a pipeline is reloaded, the rufus stop/shutdown method is called without any arguments. This instructs rufus to leave it's worker threads running. To fully shutdown the rufus worker threads, a :wait or :kill is needed. This change sends the :wait parameter to the shutdown method to ensure that the work thread is stopped. Prior to this change, the thread and all associated memory would stay in the JVM indefinitely.

Fixes #255

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 4.3.3
+  - [#255](https://github.com/logstash-plugins/logstash-input-jdbc/issues/255) Fix thread and memory leak.
+  
 ## 4.3.2
   - [#251](https://github.com/logstash-plugins/logstash-input-jdbc/issues/251) Fix connection and memory leak.
   
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -260,7 +260,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
   def stop
     close_jdbc_connection
-    @scheduler.stop if @scheduler
+    @scheduler.shutdown(:wait) if @scheduler
   end
 
   private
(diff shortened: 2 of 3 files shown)


jakelandis:
Note to reviewer : [stop](https://github.com/jmettraux/rufus-scheduler/blob/master/lib/rufus/scheduler.rb#L133) is an alias to [shutdown](https://github.com/jmettraux/rufus-scheduler/blob/master/lib/rufus/scheduler.rb#L114) and the code there is pretty clear on what it does with the options. 

I also have evidence of fix on #255

jakelandis:
@yaauie - thanks

PR #257, opened 2017-12-28
Fix issue where `sql_last_value` could lose precision

yaauie:
- DRY up the code that is tracking `sql_last_value`
 - when coercing a `sql_last_value` that is a `DateTime`, use `DateTime#to_time` to avoid losing precision


guyboertje:
merging, as I need these changes to base another change on.

PR #260, opened 2018-02-14
Force all usage of sql_last_value to be typed according to the settings

### LINKED ISSUE #140: JDBC Input set jdbc_default_timezone loses milliseconds
The bug:
1. When setting the timezone for UTC within JDBC input plug, it will lose the millesconds values of the datetime2 field ingested from sql.

For example: 2016-06-07T10:20:05.223Z with no timezone set will convert via the machine's timezone with no data loss.  e.g. if the computer was New York time, it would convert by adding +4 to the hours, and maintaining the rest of the data.
However, if you specify that the timezone is UTC by putting jdbc_default_timezone => "UTC" in your input JDBC plugin, it will convert the time to be 2016-06-07T10:20:05.000Z

Setup Info:
Logstash Version: 2.3.2 (current)
OS: Windows 7
Config File:

```
input {
  jdbc {
    jdbc_driver_library => "xxx\sqljdbc4.jar"
    jdbc_driver_class => "com.microsoft.sqlserver.jdbc.SQLServerDriver"
    jdbc_connection_string => "xxx"
    jdbc_user => "xxx"
    jdbc_password => "xxx"
    statement => "xxx"
    type => "xxx"
    jdbc_default_timezone => "UTC"
  }


}

filter {
}

output {
    stdout {
      codec => json_lines
    }
    elasticsearch{
        index => "alarms"
        document_type => "alarm"            
    }
}
```

Sample Data:
SQL, pulling data from a datetime2 field.

Steps to reproduce: Ingest data into elasticsearch via the minimal logstash config shown above using the JDBC plugin with the JDBC driver sqljdbc4.jar from a SQL database table that contains a datetime2(7) field.

guyboertje:
Added three more tests to verify the jdbc_default_timezone behaviour
when no tracking column is specified
when a numeric column is specified
when a timestamp column is specified

Fixes #140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,20 @@
+## 4.3.5
+  - [#140](https://github.com/logstash-plugins/logstash-input-jdbc/issues/140) Fix long standing bug where setting jdbc_default_timezone loses milliseconds. Force all usage of sql_last_value to be typed according to the settings.
+
 ## 4.3.4
   - [#261](https://github.com/logstash-plugins/logstash-input-jdbc/issues/261) Fix memory leak.
-  
+
 ## 4.3.3
   - [#255](https://github.com/logstash-plugins/logstash-input-jdbc/issues/255) Fix thread and memory leak.
-  
+
 ## 4.3.2
   - [#251](https://github.com/logstash-plugins/logstash-input-jdbc/issues/251) Fix connection and memory leak.
-  
+
 ## 4.3.1
   - Update gemspec summary
 
-## 4.3.0  
-  - [#147](https://github.com/logstash-plugins/logstash-input-jdbc/issues/147) Open and close connection for each query 
+## 4.3.0
+  - [#147](https://github.com/logstash-plugins/logstash-input-jdbc/issues/147) Open and close connection for each query
 
 ## 4.2.4
   - [#220](https://github.com/logstash-plugins/logstash-input-jdbc/issues/220) Log exception when database connection test fails
--- a/lib/logstash/inputs/jdbc.rb
+++ b/lib/logstash/inputs/jdbc.rb
@@ -2,7 +2,7 @@
 require "logstash/inputs/base"
 require "logstash/namespace"
 require "logstash/plugin_mixins/jdbc"
-require "yaml" # persistence
+
 
 # This plugin was created as a way to ingest data from any database
 # with a JDBC interface into Logstash. You can periodically schedule ingestion
@@ -206,21 +206,16 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
     require "rufus/scheduler"
     prepare_jdbc_connection
 
-    # Raise an error if @use_column_value is true, but no @tracking_column is set
     if @use_column_value
+      # Raise an error if @use_column_value is true, but no @tracking_column is set
       if @tracking_column.nil?
         raise(LogStash::ConfigurationError, "Must set :tracking_column if :use_column_value is true.")
       end
     end
 
-    @enable_encoding = !@charset.nil? || !@columns_charset.empty?
+    @value_tracker = LogStash::PluginMixins::ValueTracking.build_last_value_tracker(self)
 
-    # load sql_last_value from file if exists
-    if @clean_run && File.exist?(@last_run_metadata_path)
-      File.delete(@last_run_metadata_path)
-    elsif File.exist?(@last_run_metadata_path)
-      @sql_last_value = YAML.load(File.read(@last_run_metadata_path))
-    end
+    @enable_encoding = !@charset.nil? || !@columns_charset.empty?
 
     unless @statement.nil? ^ @statement_filepath.nil?
       raise(LogStash::ConfigurationError, "Must set either :statement or :statement_filepath. Only one may be set at a time.")
@@ -248,13 +243,11 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
       @scheduler = Rufus::Scheduler.new(:max_work_threads => 1)
       @scheduler.cron @schedule do
         execute_query(queue)
-        update_state_file
       end
 
       @scheduler.join
     else
       execute_query(queue)
-      update_state_file
     end
   end # def run
 
@@ -267,7 +260,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
 
   def execute_query(queue)
     # update default parameters
-    @parameters['sql_last_value'] = @sql_last_value
+    @parameters['sql_last_value'] = @value_tracker.value
     execute_statement(@statement, @parameters) do |row|
       if enable_encoding?
         ## do the necessary conversions to string elements
@@ -277,12 +270,7 @@ class LogStash::Inputs::Jdbc < LogStash::Inputs::Base
       decorate(event)
       queue << event
     end
-  end
-
-  def update_state_file
-    if @record_last_run
-      File.write(@last_run_metadata_path, YAML.dump(@sql_last_value))
-    end
+    @value_tracker.write
   end
 
   private
--- a/lib/logstash/plugin_mixins/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc.rb
@@ -3,6 +3,7 @@
 require "logstash/config/mixin"
 require "time"
 require "date"
+require "logstash/plugin_mixins/value_tracking"
 
 java_import java.util.concurrent.locks.ReentrantLock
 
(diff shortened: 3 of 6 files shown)


guyboertje:
Will rebase.

jakelandis:
> Will rebase.

sorry, i sniped `4.3.4`

guyboertje:
Did you publish 4.3.4? Can I piggy back on 4.3.4?

jakelandis:
@guyboertje - yeah, `4.3.4` is published.

jakelandis:
@guyboertje - I am trying to manually test a scenario that fails with out this change, then works with this change. Could you help me to better understand exactly how to test this ?

What I have done so far: 

Using MySQL
```
DROP TABLE test;
CREATE TABLE test (num INTEGER, update_dt_tm DATETIME(6));
INSERT into test (num, update_dt_tm) VALUES (96, NOW(6));
INSERT into test (num, update_dt_tm) VALUES (97, NOW(6));
INSERT into test (num, update_dt_tm) VALUES (98, NOW(6));
INSERT into test (num, update_dt_tm) VALUES (99, NOW(6));
SELECT * from test;
```
and the following config:
```
input {
  jdbc {
    jdbc_driver_library => "/Users/jake/workspace/examples/sql/mysql-connector-java-5.1.44/mysql-connector-java-5.1.44-bin.jar"
    jdbc_driver_class => "com.mysql.jdbc.Driver"
    jdbc_connection_string => "jdbc:mysql://localhost:3306/test2"
    jdbc_user => "XX"
    jdbc_password => "XX"
    schedule => "* * * * * *"
    statement => "SELECT * from test LIMIT 10"
    jdbc_default_timezone => "UTC"
#    record_last_run => true 
#    last_run_metadata_path => "/tmp/logstash_jdbc_last_run_1"
    }
}
output {
		stdout { codec => rubydebug}
}
```

Tried different `TIMESTAMP` and `DATETIME` types and various jdbc_default_timezones and looked for truncation of milliseconds in the output or in the last run file.  I have read through https://github.com/logstash-plugins/logstash-input-jdbc/issues/140, but don't think I am understanding the issue that this PR is addressing.

guyboertje:
I will sent an email with a link to a SFDC case that describes the fault better.

The original code for timezone support was committed in Dec 2015 when `Time.parse(sql_last_value.to_s)` was used if the tracked column value was a DateTime (converted from sql.Timestamp by Sequel).

@yaauie very recently changed this in master to `sql_last_value.to_time`, this in itself does the preserve sub-seconds.

I did my investigations before the above change. However even with the above change in place Sequel does not perform the timezone adjustments correctly. This PR refactors to allow Sequel to do the TZ adjustments correctly **while preserving Ry's intended fix**. 

I backported the new test at line into **master**.
**Note: the test has `ENV["TZ"] = "Etc/UTC"`.**

The timestamps added to the DB reflects the Singapore timezone...
`db[:test1_table].insert(:num => 22, :custom_time => "2015-01-01 20:00:00.811", :created_at => Time.now.utc)` i.e. Jan 1, 20h00

This is what I see when I add print `query.sql`:
Between the queries below, before Sequel converts it, the value of `custom_time` of the last record is '2015-01-01 20:00:00.722000'

With Ry's change reverted.
```
SELECT * from test1_table WHERE custom_time > '1970-01-01 00:00:00.000000' ORDER BY custom_time

SELECT * from test1_table WHERE custom_time > '2015-01-02 02:00:00.000000' ORDER BY custom_time
```
NOTE: sub-seconds NOT preserved AND timestamp is UTC (+600)

With Ry's change.
```
SELECT * from test1_table WHERE custom_time > '1970-01-01 00:00:00.000000' ORDER BY custom_time

SELECT * from test1_table WHERE custom_time > '2015-01-02 02:00:00.722000' ORDER BY custom_time
```
NOTE: sub-seconds IS preserved BUT timestamp is UTC (+600).

With the PR:
```
SELECT * from test1_table WHERE custom_time > '1999-12-31 18:00:00.987000' ORDER BY custom_time

SELECT * from test1_table WHERE custom_time > '2015-01-01 20:00:00.722000' ORDER BY custom_time
```
NOTE: sub-seconds IS preserved BUT timestamp is 6 hours into the future.

How to recreate with LS v5.5.2 (sub-seconds zeroed out):
Logstash runs in UTC TZ with jdbc_input settings:
"statement" => "SELECT * from test WHERE added_at > :sql_last_value ORDER BY added_at",
"use_column_value" => true,
"tracking_column" => "added_at",
"tracking_column_type" => "timestamp"
"jdbc_default_timezone" => "America/Chicago"

Add 1 record with a timestamp column (called say, `added_at`) set to 5:55 in Chicago - 5 minutes before noon UTC.
Run LS with:
You should see one event.
The sql_last_value substituted in the query should be UNIX epoch.
Persisted sql_last_value should be the `added_at` value in UTC "11:55".
Event added_at field should be converted to UTC - "11:55"

Add 1 record with `added_at` set to 6:05 - 5 minutes after noon UTC.
Run LS again.
You should see no events
The sql_last_value substituted in the query should be `added_at` value in UTC but no records exist for > "11:55".

Delete the persisted sql last value and truncate the test table
Switch to using a timezone 6 hours later than UTC say, "Asia/Almaty".
"jdbc_default_timezone" => "Asia/Almaty"

Add 1 record with `added_at`) set to "17:55"- 5 minutes before noon UTC.
Run LS with:
You should see one event.
The sql_last_value substituted in the query should be UNIX epoch.
Persisted sql_last_value should be the `added_at` value in UTC "11:55".
Event added_at field should be converted to UTC "11:55".

Add 1 record with `added_at` set to "18:55" - 5 minutes after noon UTC.
Run LS again.
You should see 2 events
The sql_last_value substituted in the query should be `added_at` value in UTC but 2 records exist for > "11:55".

Try LS 6.2.1 (sub-seconds preserved, but same results)

guyboertje:
The above should hold true when when not using a tracking column and sql_last_value is the time of run in UTC but the DB times are not UTC, i.e.:
"statement" => "SELECT * from test WHERE added_at > :sql_last_value ORDER BY added_at",
"tracking_column_type" => "timestamp"
"jdbc_default_timezone" => "America/Chicago"

jakelandis:
@guyboertje - thanks for the explanation, I think I get the changes here now. 

I confirmed that the timezone issue is fixed, but w/r/t the millisecond precision, it seems that the persisted datetime/time does indeed have the milliseconds, but when the `:sql_last_value` is sent to the database does not. 

For example the persisted timestamp is `--- 2018-02-23 17:28:44.189915000 -06:00` , but the following is what is sent to the database:  `SELECT * from test WHERE update_dt_tm > '2018-02-23 17:28:43'`  

**Is this just an artifact of sequel support and my testing (MySQL) ?** 

-----------------

Also, 
While trying to understand the changes here I wrote what could be a commit log describing the change (is this correct ?): 

When  `:sql_last_value` is sent the to database, it is converted to the JVM default timezone.  This change honors the `jdbc_default_timezone` for use in the `:sql_last_value` timezone conversion. 

Assuming `America/Chicago` as the default JVM timezone,  but the configuration declares `UTC` as the database timezone `jdbc_default_timezone => "UTC"`  
The last_run is correctly recorded as `2018-02-23 22:30:34.054592000 Z` , but when the `> :sql_last_value` gets sent to the database, it incorrectly gets converted to the JVM default timezone ( ` > '2018-02-23 16:30:34'`  ).  

This change fixes the conversion when sending the `:sql_last_value` to the database to honor the `jdbc_default_timezone` setting.  Specifically, this leverages the fact that the Sequel library will handle the timezone conversions properly if passed a DateTime object, and won't don't any timezone conversions if passed a Time object.  

This change also refactors parts of the code for better readability and adds more tests.

jakelandis:
> Is this just an artifact of sequel support and my testing (MySQL) ?

Answering my own question ... I dug into sequel a bit and found this magic is needed to enable fractional seconds on mysql 5.6.5+

```
  sequel_opts => {
       fractional_seconds => true
     }
```
The ms now show up
```
 SELECT * from test WHERE update_dt_tm > '2018-02-24 00:26:06.226550'
```

yangzhiwen911:
> > Is this just an artifact of sequel support and my testing (MySQL) ?
> 
> Answering my own question ... I dug into sequel a bit and found this magic is needed to enable fractional seconds on mysql 5.6.5+
> 
> ```
>   sequel_opts => {
>        fractional_seconds => true
>      }
> ```
> 
> The ms now show up
> 
> ```
>  SELECT * from test WHERE update_dt_tm > '2018-02-24 00:26:06.226550'
> ```

where did u  add     sequel_opts => {
       fractional_seconds => true
     }      to？？

Issue #274, opened 2018-03-22
sql_last_value can result in NULL for the SQL query


jakelandis:
This is a regression introduced in `4.3.5` of the this plugin. 

Specific Date/Timestamp values may result in a valid value stored in the `last_run_metadata_path`, but not recognized correctly when read.  This can result in a `NULL` passed to the SQL query. 

Specifically the following format in the `last_run_metadata_path`
```
--- '2010-03-19T14:48:40.483Z'
```
can result in 
```
SELECT update_dt_tm as update_dt_tm from test WHERE update_dt_tm > NULL  ORDER BY update_dt_tm asc LIMIT 1
```

...but a format like this:
```
--- 2018-02-23 16:48:15.339130000 -06:00
```
will work as expected. 

The root cause here is the duck typing of the Ruby YAML parser we use to read this value. In the former case (the NULL), the value quacks like a String, and the latter case (working) it quacks like Ruby Time object. 

The fix here is to explicitly parse to Time and DateTime types if they don't implicitly get read as that type.

PR #275, opened 2018-03-22
 Fix regression with 4.3.5 that can result in NULL :sql_last_value depending on timestamp forma

jakelandis:
The specs without the code fix illustrates the issue. 

There is also a minor fix to the tests included in this PR to reset Timecop (it was causing intermittent failures of the new specs).

### DIFF
--- a/spec/inputs/jdbc_spec.rb
+++ b/spec/inputs/jdbc_spec.rb
@@ -739,6 +739,65 @@ describe LogStash::Inputs::Jdbc do
     end
   end
 
+  context "when previous runs are to be respected upon successful query execution (by time string)" do
+
+    let(:settings) do
+      { "statement" => "SELECT custom_time FROM test_table WHERE custom_time > :sql_last_value",
+        "use_column_value" => true,
+        "tracking_column" => "custom_time",
+        "tracking_column_type" => "timestamp",
+        "last_run_metadata_path" => Stud::Temporary.pathname }
+    end
+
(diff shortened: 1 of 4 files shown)


spec/inputs/jdbc_spec.rb
```diff
@@ -410,6 +412,7 @@
         event = queue.pop
         expect(event.get("num")).to eq(12)
         expect(event.get("custom_time").time).to eq(Time.iso8601("2015-01-02T03:21:21.000Z"))
+        Timecop.return
```
guyboertje: When I added this test I took care to use the block form of `Timecop.travel` because the docs said it does the return after the block yield.

Did you verify that this particular `Timecop.return` was necessary? If so, we should raise an issue in the Timecop library.

guyboertje:
So how does this manifest?
Is it only when the `last_run_metadata_path` file exists with a string value before the update to the new version or when plugin is used in a never run before scenario?

spec/inputs/jdbc_spec.rb
jakelandis: This one was not the issue... i just did a quick search for Timecop without a return. I will revert this line.

jakelandis:
> Is it only when the last_run_metadata_path file exists with a string value before the update to the new version or when plugin is used in a never run before scenario?

Yes, upgrading from a prior version AND the time stamp is in a format like referenced on #274 (which is a valid time stamp, and will parse correctly .. but when read from YAML it is still a String)

Issue #273, opened 2018-03-22
Question on Use of Use_Column_Value And Timestamps 


MorrieAtElastic:
Please post all product and debugging questions on our [forum](https://discuss.elastic.co/c/logstash). Your questions will reach our wider community members there, and if we confirm that there is a bug, then we can open a new issue here.

For all general issues, please provide the following details for fast resolution:

- Version: 5.6.4 Customer; 5.6.5 reproduced in house against Oracle V12
- Operating System: RHEL
- Config File (if you have sensitive info, please remove it):(included below)
- Sample Data: n/a 
- Steps to Reproduce: 

- When using the JDBC Plugin to query Oracle databases, it seems that USE_COLUMN_VALUE must be set to true even when using timestamp fields as the tracking column. This is inconsistent and completely opposite to the documented discussion of USE_COLUMN_VALUE which states that it should be used only when the tracking column is not a timestamp column. If USE_COLUMN_VALUE is set to false, timestamp comparisons do not work correctly and documents which should be issued to elasticsearch are not. See https://www.elastic.co/guide/en/logstash/5.6/plugins-inputs-jdbc.html#plugins-inputs-jdbc-use_column_value for the documented definition of USE_COLUMN_VALUE ("set to true when tracking column is not a timestamp"). 

- A customer discovered this, provided a test case which I have reproduced his results; the problem can be demonstrated without connecting to an Elasticsearch cluster,  just by using the rubydebug output. 

*Test configuration file:* 

```
input {
    jdbc {
        jdbc_validate_connection => true
        jdbc_driver_library => "/u01/app/oracle/product/12.2/db_1/md/property_graph/lib/ojdbc7.jar"
        jdbc_driver_class => "Java::oracle.jdbc.driver.OracleDriver"
        jdbc_connection_string => "jdbc:oracle:thin:@//localhost:1521/orcl12c"
        jdbc_user => "system"
        jdbc_password => "oracle"
        statement => "SELECT CAST(add_months(sysdate, -1) AS TIMESTAMP) AS time_column, 'Some Value' AS some_data FROM dual WHERE CAST(add_months(sysdate, -1) AS TIMESTAMP) > :sql_last_value"
        use_column_value => true
        tracking_column => "time_column"
        tracking_column_type => "timestamp"
        schedule => "*/5 * * * * *"
        record_last_run =>  true
        tags => ["columnvalue"]
[...]

PR #277, opened 2018-03-29
Clarify use of use_column_value

karenzone:
Clarify use of use_column_value. 
Make last_run_metadata_path reference in record_last_run entry clickable.

Phase 1 fix for #273.
Phase 2 will involve reworking the topic

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -100,19 +100,19 @@ option is typically used when the SQL statement is large or cumbersome to supply
 The file option only supports one SQL statement. The plugin will only accept one of the options.
 It cannot read a statement from a file as well as from the `statement` configuration parameter.
 
-==== Configuring multiple SQL statements 
+==== Configuring multiple SQL statements
 
-Configuring multiple SQL statements is useful when there is a need to query and ingest data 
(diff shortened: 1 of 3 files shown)


docs/index.asciidoc
```diff
@@ -449,38 +449,39 @@ here, ":target_id" is a named parameter. You can configure named parameters
...
 
 [id="plugins-{type}s-{plugin}-tracking_column"]
-===== `tracking_column` 
+===== `tracking_column`
 
   * Value type is <<string,string>>
   * There is no default value for this setting.
 
-If tracking column value rather than timestamp, the column whose value is to be tracked
+The column whose value is to be tracked (if tracking column value rather than the last time the query was run)
```
jakelandis: How about "The column whose value is to be tracked if use_column_value is set to true" ?

jakelandis:
@karenzone - changes look good.  one minor suggestion. 

For phase 2, I will log a proposal issue. I think we can get rid of one (maybe two of these settings).

jakelandis:
Logged https://github.com/logstash-plugins/logstash-input-jdbc/issues/279 for phase 2

karenzone:
Yes, I tested without issue

PR #291, opened 2018-07-19
connect our logger up to the charset converters

yaauie:
The `LogStash::Util::Charset` utility will attempt to send a `warn` to its
`@logger` instance variable when it encounters invalid encoding, but does not
take a logger to its constructor.

Failing to wire up a logger means that each time we attempt to convert a string
and fail to arrive at valid UTF-8, we send `warn` to `NilClass`, which raises
a `NoMethodError`.


danhermann (approved):
LGTM

manzoor77:
An unknown error occurred sending a bulk request to Elasticsearch (will retry indefinitely) {:message=>""\xFF" from ASCII-8BIT to UTF-8", :exception=>LogStash::Json::GeneratorError, :backtrace=>["/Users/manzoorfaisal/Desktop/Laptop-Migration-2/logstash-7.17.4/logstash-core/lib/logstash/json.rb:43:in 

can you please me to out from this issue?

PR #294, opened 2018-08-13
Add transparent support for count query embargo on failure.

guyboertje:
Many folks get errors when using DB drivers that are not supported by adapters in Sequel.

This update traps any errors when doing the first count query and prevents further count queries from being executed.


guyboertje:
Changelog and gemspec updates to follow.

lib/logstash/plugin_mixins/checked_count_logger.rb
```diff
@@ -0,0 +1,35 @@
+
+module LogStash module PluginMixins
```
jsvd: IMO we should not place plugin specific (or integration specific) mixins in `LogStash::PluginMixins`, all of these could be namespaced to `LogStash::PluginMixins::Jdbc` or similar

lib/logstash/plugin_mixins/checked_count_logger.rb
```diff
@@ -0,0 +1,35 @@
+
```
jsvd: missing encoding header

lib/logstash/plugin_mixins/checked_count_logger.rb
```diff
@@ -0,0 +1,35 @@
...
+      else
+        @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters)
+      end
+    end
+
+    def check_count_query(query)
+      begin
+        execute_count(query)
+        @count_is_supported = true
+      rescue Exception => e
```
jsvd: is there a risk that the query fails here for an arbitrary reason not related to counts?

lib/logstash/plugin_mixins/checked_count_logger.rb
```diff
@@ -0,0 +1,35 @@
...
+      @needs_check = true
+      @count_is_supported = false
+      @in_debug = @logger.debug?
+    end
+
+    def log_statement_parameters(query, statement, parameters)
+      return unless @in_debug
+      check_count_query(query) if @needs_check
+      if @count_is_supported
+        @logger.debug("Executing JDBC query", :statement => statement, :parameters => parameters, :count => query.count)
```
jsvd: since `execute_count` was implemented to do `query.count`, maybe we use it here as well?

spec/inputs/jdbc_spec.rb
```diff
@@ -1267,4 +1267,40 @@
...
+    after do
+      plugin.stop
+    end
+
+    it "should not log a debug line with a count key" do
+      expect(logger).to receive(:info).once.with("Disabling count queries as the executing the count SQL raised an error")
+      expect(logger).to receive(:debug).once.with("Executing JDBC query", :statement => settings["statement"], :parameters => {:sql_last_value=>"bar"})
+      plugin.run(queue)
+      event = queue.pop
+      expect(event.get("num")).to eq(1)
```
jsvd: this test is about not debugging with a count key, maybe we can remove the remaining assertions and create a separate test to check the presence of things that we expect?

lib/logstash/plugin_mixins/checked_count_logger.rb
guyboertje: I suppose so. I'll see if I can find a more specific Exception to rescue.

spec/inputs/jdbc_spec.rb
guyboertje: Sure. I can remove the other expectations, as this context block is a copy of the one at line 1240, those expectations have been verified already.
OTOH, it could be argued that it is important to verify that the normal event creation still occurs when the code from this PR is in the call path and when some degree of stubbing is used.

jsvd: Agreed it may be desirable to test that but I'd put it into a separate `it` block that uses the same test setup (thus it is the same scenario we're testing)

guyboertje:
@jsvd 

Changelog and gemspec updates made. Is this PR OK now?

PR #304, opened 2018-10-12
Give ability poll a source system more than once per minute.

robbruce:
The rufus scheduler supports a schedule based on seconds, by using the `repeat` method instead of `cron`, the method returns an EveryJob or a CronJob based on the schedule used.

```
schedule => "* * * * *" # CronJob scheduled for every minute
```

```
schedule => "10" # EveryJob scheduled for every 10 seconds
```


robbruce:
I have signed the CLA at a company level, twice! Please can this be checked.

rverchere:
You should add a parameter `schedule_type` (values: cron or repeat) to avoid regressions.

deslauriersp:
> The rufus scheduler supports a schedule based on seconds,

Rufus supports seonds, and Rufus support cron "**like**", which means it looks like cron, but it isn't cron. Rufus add the "seconds" support, as per their doc. Notice the extra "*"

```
"* * * * * *" //should be every seconds
"*/10 * * * * *" //every ten seconds (works in my pipeline logstash 7.6)
```

robbruce:
Closing as although CLA has been signed, it's just being picked up

PR #315, opened 2018-11-30
Add zookeeper as last value meta path storage

VladislavRusin:
Added feature of getting and saving last value in zookeeper.
It will cover following cases:
* When we have "Active - Stand by" architecture where 2 logstash instances work on different machines and they need to use same metadata path
* When we need to refill old events by changing last value (In case of file system it's difficult to change values)


elasticcla:
Hi @VladislavRusin, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/logstash-plugins/logstash-input-jdbc/pull/315.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #330, opened 2019-03-20
Allow to stop JDBC input if there is a huge number of rows

prehor:
If a huge number of rows is processed, JDBC input continuously reads rows from the database and cannot be stopped. When Logstash is killed, it does not remember the last processed row and starts again from the beginning.


prehor:
Hello @jsvd. I think that the problem is still not fixed. If you are interested, I can send PR for the current version. P.

PR #343, opened 2019-07-17
Jdk11 class loading

guyboertje:
This is the implementation of a wrapped class loader as suggested by Dan.

There is a small code re-org and due to a Jruby bug in 9.2.X, a monkey patch to tzinfo.

Gemspec and docs changes to follow.


lib/logstash/inputs/tzinfo_jruby_patch.rb
```diff
@@ -0,0 +1,39 @@
...
+    # A time represented as an `Integer` number of seconds since 1970-01-01
+    # 00:00:00 UTC (ignoring leap seconds), the fraction through the second
+    # (sub_second as a `Rational`) and an optional UTC offset. Like Ruby's `Time`
+    # class, {Timestamp} can distinguish between a local time with a zero offset
+    # and a time specified explicitly as UTC.
+    class Timestamp
+
+      protected
+
+      def new_datetime(klass = DateTime)
```
colinsurprenant: Is this a monkey patch of an existing method or is this a new method added? This should be documented. If this is a new method, is there a specific reason to add it in TZInfo::Timestamp?

lib/logstash/inputs/jdbc.rb
```diff
@@ -2,7 +2,7 @@
 require "logstash/inputs/base"
 require "logstash/namespace"
 require "logstash/plugin_mixins/jdbc/jdbc"
-
+require_relative "tzinfo_jruby_patch"
```
colinsurprenant: related to my other comment, if this is a required monkey patch, is should be guarded by version checks to make sure it does not blow up in a future update.

lib/logstash/inputs/tzinfo_jruby_patch.rb
guyboertje: It is a patch on the TZInfo::Timestamp class, I'm redefining the `new_datetime` method.

lib/logstash/inputs/jdbc.rb
guyboertje: Do you mean guard on the JRuby version > 9.0.0.0 and < 9.2.8.0?
Headius tagged the issue with 9.2.8.0 but there is no guarantee.

colinsurprenant:
Per my above code comments, I'd like for monkey patches to be documented, that is, if a method is redefined it should be documented as such; future developers will then directly know without having do dig in the original class to figure that out. Also, I'd like the inclusion of a monkey patch to be guarded by a version check; any future upgrade of JRuby will potentially conflicts with this monkey patch and we need an explicit way of forcing that verification at build time. The version check can simply fail the build if the version is different than the version upon which that monkey patch was built for.

guyboertje:
It occurs to me that due to BWC the patch can only really be removed when the build matrix is has all versions using JRuby 9.2.8.0 and newer.
The build failure I have now will act as a reminder to circle back to revisit this as and when the build matrix fails the 7.X and 8.X SNAPSHOT lines.

colinsurprenant:
this will generate an exception at runtime wouldn't be better to crash in the tests/specs only?

guyboertje:
It will crash the specs locally and fail the build in travis, which should prevent a plugin gem being released until the checks have been done.

Would you prefer in the spec file?:
```
RSpec.configure do |c|
  tzinfo_jruby_bugfixed_version = "9.2.8.0"
  current_jruby_version = Gem::Version.new(JRUBY_VERSION)
  patched_jruby_version = Gem::Version.new(tzinfo_jruby_bugfixed_version)
  tzinfo_patch_remove_message = "The TZInfo patch file must be removed, the JRuby bug that this patch provides has been fixed in #{tzinfo_jruby_bugfixed_version} (but please check)"
  raise tzinfo_patch_remove_message unless current_jruby_version < patched_jruby_version
end
```

lib/logstash/inputs/tzinfo_jruby_patch.rb
```diff
@@ -0,0 +1,57 @@
...
+# Java Long. There is no appreciable precision loss at the microsecond level
+
+tzinfo_jruby_bugfixed_version = "9.2.8.0"
+tzinfo_jruby_bugadded_version = "9.2.0.0"
+
+current_jruby_version = Gem::Version.new(JRUBY_VERSION)
+broken_jruby_version = Gem::Version.new(tzinfo_jruby_bugadded_version)
+patched_jruby_version = Gem::Version.new(tzinfo_jruby_bugfixed_version)
+
+return unless current_jruby_version >= broken_jruby_version && current_jruby_version < patched_jruby_version
```
colinsurprenant: 👍

PR #348, opened 2019-09-03
Upd changelog + ensure drivers loaded once (omission from previous PR)

guyboertje:
Towards the end of the to-ing and fro-ing of the last PR, I noticed the driver loading code was being loaded for each scheduled invocation but I neglected to add the change to the PR.


colinsurprenant:
LGTM

PR #349, opened 2019-09-05
Add Prepared Statement support

guyboertje:
Code, tests, docs and version bump.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -140,6 +140,35 @@ input {
 }
 ---------------------------------------------------------------------------------------------------
 
+==== Prepared Statements
+
+Using server side prepared statements can speed up execution times as the server optimises the query plan and execution.
+
+NOTE: Not all JDBC accessible technologies will support prepared statements.
+
+With the introduction of Prepared Statement support comes a different code execution path and some new settings. Most of the existing settings are still useful but there are several new settings for Prepared Statements to read up on.
+Use the boolean setting `use_prepared_statements` to enable this execution mode. Use the `prepared_statement_name` setting to specify a name for the Prepared Statement, this identifies the prepared statement locally and remotely and it should be unique in your config and on the database. Use the `prepared_statement_bind_values` array setting to specify the bind values, use the exact string `:sql_last_value` (multiple times if necessary) for the predefined parameter mentioned before. The `statement` (or `statement_path`) setting still holds the SQL statement but to use bind variables you must use the `?` character as a placeholder in the exact order found in the `prepared_statement_bind_values` array.
+
+NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
+
+Example:
+[source,ruby]
+---------------------------------------------------------------------------------------------------
+input {
+  jdbc {
+    statement => "SELECT * FROM mgd.seq_sequence WHERE _sequence_key > ? AND _sequence_key < ? + ? ORDER BY _sequence_key ASC"
+    prepared_statement_bind_values => [":sql_last_value", ":sql_last_value", 4]
+    prepared_statement_name => "foobar"
+    use_prepared_statements => true
+    use_column_value => true
+    tracking_column_type => "numeric"
+    tracking_column => "_sequence_key"
+    last_run_metadata_path => "/elastic/tmp/testing/confs/test-jdbc-int-sql_last_value.yml"
+    # ... other configuration bits
+  }
+}
+---------------------------------------------------------------------------------------------------
+
 
 [id="plugins-{type}s-{plugin}-options"]
 ==== Jdbc Input Configuration Options
@@ -149,7 +178,6 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 [cols="<,<,<",options="header",]
 |=======================================================================
 |Setting |Input type|Required
-| <<plugins-{type}s-{plugin}-plugin_timezone>> |<<string,string>>, one of `["local", "utc"]`|No
 | <<plugins-{type}s-{plugin}-clean_run>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-columns_charset>> |<<hash,hash>>|No
 | <<plugins-{type}s-{plugin}-connection_retry_attempts>> |<<number,number>>|No
@@ -170,6 +198,9 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 | <<plugins-{type}s-{plugin}-last_run_metadata_path>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-lowercase_column_names>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-parameters>> |<<hash,hash>>|No
+| <<plugins-{type}s-{plugin}-plugin_timezone>> |<<string,string>>, one of `["local", "utc"]`|No
+| <<plugins-{type}s-{plugin}-prepared_statement_bind_values>> |<<array,array>>|No
+| <<plugins-{type}s-{plugin}-prepared_statement_name>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-record_last_run>> |<<boolean,boolean>>|No
 | <<plugins-{type}s-{plugin}-schedule>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-sequel_opts>> |<<hash,hash>>|No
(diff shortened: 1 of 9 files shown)


lib/logstash/inputs/jdbc.rb
```diff
@@ -201,6 +201,12 @@ module LogStash module Inputs class Jdbc < LogStash::Inputs::Base
   # this will only convert column0 that has ISO-8859-1 as an original encoding.
   config :columns_charset, :validate => :hash, :default => {}
 
+  config :use_prepared_statements, :validate => :boolean, :default => false
+
+  config :prepared_statement_name, :validate => :string, :default => SecureRandom.hex(40)
```
colinsurprenant: could this potentially lead into a "prepared statements" memory leak on the database if the default is used and the plugin is repetitively restarted?

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -0,0 +1,126 @@
+# encoding: utf-8
+
+module LogStash module PluginMixins module Jdbc
+  class StatementHandler
+    def self.build_statement_handler(plugin, logger)
+      klass = plugin.use_prepared_statements ? PreparedStatementHandler: NormalStatementHandler
```
colinsurprenant:
```suggestion
      klass = plugin.use_prepared_statements ? PreparedStatementHandler : NormalStatementHandler
```

logstash-input-jdbc.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name = 'logstash-input-jdbc'
-  s.version         = '4.3.15'
+  s.version         = '4.3.16'
```
colinsurprenant: IIUC this is a new feature with proper defaults to stay BWC which allows a patch release?

colinsurprenant:
Great work @guyboertje  - left a few minor comments but otherwise LGTM!

docs/index.asciidoc
```diff
@@ -140,6 +140,35 @@ input {
...
+Use the boolean setting `use_prepared_statements` to enable this execution mode. Use the `prepared_statement_name` setting to specify a name for the Prepared Statement, this identifies the prepared statement locally and remotely and it should be unique in your config and on the database. Use the `prepared_statement_bind_values` array setting to specify the bind values, use the exact string `:sql_last_value` (multiple times if necessary) for the predefined parameter mentioned before. The `statement` (or `statement_path`) setting still holds the SQL statement but to use bind variables you must use the `?` character as a placeholder in the exact order found in the `prepared_statement_bind_values` array.
+
+NOTE: Building count queries around a prepared statement is not supported at this time and because jdbc paging uses count queries under the hood, jdbc paging is not supported with prepared statements at this time either. Therefore, `jdbc_paging_enabled`, `jdbc_page_size` settings are ignored when using prepared statements.
```
robbavey: Do you think it makes sense to actively warn users or even fail to start if that use these settings that they are ignored when using prepared statements? Or mention in the definitions of these values that they are ignored when `use_prepared_statement` is false

docs/index.asciidoc
```diff
@@ -409,6 +441,22 @@ Whether to force the lowercasing of identifier fields
...
+  * Value type is <<array,array>>
+  * Default value is `[]`
+
+Array of bind values for the prepared statement. `:sql_last_value` is a reserved predefined string
+
+[id="plugins-{type}s-{plugin}-prepared_statement_name"]
+===== `prepared_statement_name`
+
+  * Value type is <<string,string>>
+  * Default value is `""`
```
robbavey: This is set to a random string in the code, rather than blank

lib/logstash/inputs/jdbc.rb
```diff
@@ -277,8 +283,9 @@ def stop
 
   def execute_query(queue)
     # update default parameters
-    @parameters['sql_last_value'] = @value_tracker.value
-    execute_statement(@statement, @parameters) do |row|
+    # @parameters['sql_last_value'] = @value_tracker.value
```
robbavey: Should these commented out statements be removed?

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
```diff
@@ -0,0 +1,126 @@
...
+      query = build_query(db, sql_last_value)
+      query.each do |row|
+        yield row
+      end
+    end
+
+    private
+
+    def build_query(db, sql_last_value)
+      # don't log statement count when using prepared statements for now...
```
robbavey: Does it make sense to log the statement, just not the count?

lib/logstash/plugin_mixins/jdbc/value_tracking.rb
```diff
@@ -5,28 +5,26 @@ module LogStash module PluginMixins module Jdbc
   class ValueTracking
 
     def self.build_last_value_tracker(plugin)
+      handler = NullFileHandler.new(plugin.last_run_metadata_path)
```
robbavey: Could we do this as a one-liner to avoid unnecessary creation of `NullFileHandler`?

```suggestion
      handler = plugin.record_last_run ? FileHandler.new(plugin.last_run_metadata_path) : NullFileHandler.new(plugin.last_run_metadata_path)
```

spec/inputs/jdbc_spec.rb
```diff
@@ -11,6 +11,10 @@
 require "time"
 require "date"
 
+# ENV["LOG_AT"].tap do |level|
```
robbavey: Should this commented out section be removed?

spec/inputs/jdbc_spec.rb
```diff
@@ -1310,4 +1314,91 @@
...
+    end
+
+    after do
+      plugin.stop
+    end
+    let(:last_run_value) { 250 }
+    let(:expected_queue_size) { 100 }
+    let(:num_rows) { 1000 }
+
+    context "without bind paramters" do
```
robbavey: Nit: typo

logstash-input-jdbc.gemspec
guyboertje: Yes, it is 100% BWC.

lib/logstash/inputs/jdbc.rb
guyboertje: You raise a good point about lifetimes and lifecycles of the Prepared Statements.
For some DB technologies, the prepared statements are associated with the connection  (session) - when the connection is closed the caches are cleansed. For others, I dunno.
Certainly, we should call `close` on the prepared statement Sequel object when the plugin stops however I can't find evidence that closing the prepared statement affects its cached info. I think we are OK using a random string in the event of a user forgetting to set the `prepared_statement_name ` setting but I'll add this hook into the plugin stop call anyway.

docs/index.asciidoc
guyboertje: Thanks, see [this comment](https://github.com/logstash-plugins/logstash-input-jdbc/pull/349#discussion_r321638743), I'm happy to leave or remove the random string and adjust the docs here depending on Colin's reply.

docs/index.asciidoc
guyboertje: > Or mention in the definitions of these values that they are ignored when use_prepared_statement is false

They are not ignored when use_prepared_statement is false


Perhaps in the near future one of you folks will add the COUNT enhancement. My gut feeling is that a separate count_statement setting will need to be added because we can't rely on Sequel generating the correct SQL to paginate prepared statements.

Given (IMO) that pagination can't work without adding the above setting and therefore can't be a transparent plugin upgrade for the user (i.e. pagination just starts working) then a hard validation fail is necessary.
/cc @colinsurprenant

lib/logstash/plugin_mixins/jdbc/statement_handler.rb
guyboertje: In a round about way I guess you are correct. The decision of what to log is a concern of the CheckedCountLogger class with a hint from the flavour of statement handler. The PreparedStatementHandler should be able to unset the `@needs_check` ivar in some way - then the logging without count will take place.

docs/index.asciidoc
robbavey: Yes, I meant `use_prepared_statement` is `true` 🤦‍♂

lib/logstash/inputs/jdbc.rb
colinsurprenant: or we make the name a required option?

guyboertje: It can't be a validation as a standard required setting because it is mode dependent. I suppose I could introduce some conditional validation.

colinsurprenant: ah I see... no need to block PR for this but it feels like a potential long term issue.

PR #352, opened 2019-09-27
Doc changes

robbavey:
Add additional troubleshooting section for jdbc driver loading and remove confusing language about mixins.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -317,8 +317,12 @@ database timezone is UTC then you do not need to set either of these settings.
   * There is no default value for this setting.
 
 JDBC driver class to load, for example, "org.apache.derby.jdbc.ClientDriver"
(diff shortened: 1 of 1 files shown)


karenzone:
Do you think the note might get more attention here:  https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html#_drivers

Maybe we can keep the note where you think best, and reference it from the other location if you think that's appropriate.

robbavey:
@karenzone Should we move the notes up to the section you suggested, or have the information repeated?

karenzone:
What about adding something like "See the
<<plugins-{type}s-{plugin}-jdbc_driver_library>> and
<<plugins-{type}s-{plugin}-jdbc_driver_class>> options for more info."` to the **Drivers** section?

PR #353, opened 2019-10-17
[doc] Add an example for sequel_opts

lucabelluccini:
Documentation addendum on the sequel_opts usage.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -485,7 +485,7 @@ exactly once.
 
 General/Vendor-specific Sequel configuration options.
 
-An example of an optional connection pool configuration:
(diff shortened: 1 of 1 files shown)


kares:
Hey Luca, the plugin is in a transition state to an integration one.
So would recommend moving the PR: https://github.com/logstash-plugins/logstash-integration-jdbc
These 2 options might also make sense to be exposed directly, instead of being under the internal `sequel_opts: ...` so you could possibly also (only) open an issue for discussion regarding these.

lucabelluccini:
Ok, I'll close the issue and open an issue on https://github.com/logstash-plugins/logstash-integration-jdbc

Thank you for the heads up. I wasn't aware of it.

Issue #355, opened 2019-10-24
Oracle: Unable to find driver class via URLClassLoader


lukasertl:
- Version:
7.4.1

- Operating System:
CentOS 7

- Config File (if you have sensitive info, please remove it):

```
    jdbc {
        jdbc_validate_connection => true
        jdbc_connection_string   => "CONN"
        jdbc_user                => "USER"
        jdbc_password            => "PW"
        jdbc_driver_library      => "/usr/lib64/oracle/19.3/client64/lib/ojdbc8.jar"
        jdbc_driver_class        => "Java::oracle.jdbc.driver.OracleDriver"
        statement                => "select ..."
        schedule                 => "* * * * *"
        last_run_metadata_path   => "/opt/logstash/jdbc_metadata/xxx"
        id                       => "input-jdbc-xxx"
        add_field                => { "database" => "xxx" }
    }
```

I've dozens of JDBC inputs like the above - all stopped working after upgrade to logstash 7.4.1 (worked fine in 7.4.0).  The Oracle driver is from Oracle Instantclient 19.3.

This is the error I get:

```
{ 2020 rufus-scheduler intercepted an error:
2020   job:
2020     Rufus::Scheduler::CronJob "* * * * *" {}
2020   error:
2020     2020
2020     LogStash::PluginLoadingError
2020     Unable to find driver class via URLClassLoader in given driver jars: Java::oracle.jdbc.driver.OracleDriver and Java::oracle.jdbc.driver.OracleDriver
2020       /usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-input-jdbc-4.3.16/lib/logstash/plugin_mixins/jdbc/jdbc.rb:181:in `open_jdbc_connection'
2020       /usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-input-jdbc-4.3.16/lib/logstash/plugin_mixins/jdbc/jdbc.rb:253:in `execute_statement'
2020       /usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-input-jdbc-4.3.16/lib/logstash/inputs/jdbc.rb:309:in `execute_query'
2020       /usr/share/logstash/vendor/bundle/jruby/2.5.0/gems/logstash-input-jdbc-4.3.16/lib/logstash/inputs/jdbc.rb:276:in `block in run'
[...]

PR #356, opened 2019-10-25
Fix driver class loading

robbavey:
The changes to class loading intended to fix usage of :jdbc_driver_library with
Java 11, introduced in `v4.3.14`, appear to have some issues
(https://github.com/elastic/logstash/issues/11268, #355). This commit attempts
to resolve those issues, and reworks the integration tests to work with
docker and ensure that the :jdbc_driver_library configuration option works.

This has been tested with postgresql using Java 8 and Java 11.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc/jdbc.rb
@@ -140,28 +140,18 @@ module LogStash  module PluginMixins module Jdbc
 
     private
 
-    def load_drivers
-      return if @jdbc_driver_library.nil? || @jdbc_driver_library.empty?
-
-      driver_jars = @jdbc_driver_library.split(",")
-
-      # Needed for JDK 11 as the DriverManager has a different ClassLoader than Logstash
-      urls = java.net.URL[driver_jars.length].new
-
-      driver_jars.each_with_index do |driver, idx|
-        urls[idx] = java.io.File.new(driver).toURI().toURL()
-      end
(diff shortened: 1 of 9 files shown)


lib/logstash/plugin_mixins/jdbc/jdbc.rb
```diff
@@ -140,28 +140,18 @@ def jdbc_connect
...
+        @jdbc_driver_library.split(",").each do |driver_jar|
+          begin
+            @logger.debug("loading #{driver_jar}")
+            # Use https://github.com/jruby/jruby/wiki/CallingJavaFromJRuby#from-jar-files to make classes from jar
+            # available
+            require driver_jar
+          rescue LoadError => e
+            raise LogStash::PluginLoadingError, "unable to load #{driver_jar} from :jdbc_driver_library, #{e.message}"
+          end
+        end
```
yaauie: can/should we also ensure that if `@jdbc_driver_class` was provided, it resolves after we've loaded the jars?

robbavey: It's effectively done [straight after](https://github.com/logstash-plugins/logstash-input-jdbc/blob/master/lib/logstash/plugin_mixins/jdbc/jdbc.rb#L178) this method is called. I'm not sure the extra check buys us a whole ton

yaauie: derp. you're right. you've got my `LGTM`

Issue #360, opened 2019-11-06
Paging bug: JDBC_PAGING_ENABLED and JDBC_PAGE_SIZE degraded after version 4.3.14


bmax:
Hello! It seems like at the same time prepared statements were introduced in version 4.3.16 is the same time paging functionality broke. I've tested this by rolling back to 4.3.14 and ensuring that `jdbc_paging_enabled` works. I am also working on a PR for a fix but in the mean time I thought I'd get this issue out first in case anyone else is having trouble.

The problem resides here: https://github.com/logstash-plugins/logstash-input-jdbc/blob/master/lib/logstash/plugin_mixins/jdbc/statement_handler.rb#L34 

The variable `@jdbc_paging_enabled` is not available in that context therefore this will never evaluate as true. 

#### Debugging information

- Version: > 4.3.14
- Operating System: centos (Docker image logstash:7.4.1)
- Config File (if you have sensitive info, please remove it):
```
input {
  jdbc {
    jdbc_driver_library => "/usr/share/logstash/bin/postgresql-42.2.8.jar"
    jdbc_driver_class => "org.postgresql.Driver"
    jdbc_connection_string => "jdbc:postgresql://${POSTGRES_URL}/${POSTGRES_DB}"
    jdbc_user => "${POSTGRES_USER}"
    jdbc_password => "${POSTGRES_PASSWORD}"
    schedule => "* * * * * *"
    jdbc_paging_enabled => true
    jdbc_page_size => 10
    statement => "SELECT * from test"
  }
}

output {
  elasticsearch {
    document_id => "visit-%{user_id}-%{merchant_id}"
    hosts    => [ "${ELASTICSEARCH_HOST}" ]
    index => "program_users"
  }
}
```
- Steps to Reproduce: Include `jdbc_paging_enabled` and ensure that the select query does not include limit or offset.

PR #361, opened 2019-11-06
Fix jdbc_paging_enabled config flag not being passed through

bmax:
fixes https://github.com/logstash-plugins/logstash-input-jdbc/issues/360

Not a ruby developer by trade so having trouble making the test better.

First PR. I signed CLA after I made this. 
cc @guyboertje


bmax:
@robbavey Any help?

robbavey:
@bmax - This has been merged and included in the latest `4.3.19` release of the JDBC input. Thanks again for your contribution!

bmax:
woohoo @robbavey! Super excited, thanks. How does the default version that gets included with Logstash's Dockerfile get updated?

robbavey:
@bmax Logstash (including the docker image) is released at the same time as the rest of the Elastic stack, and this change will be included in a future release.
However, we don't tend to give out dates and the exact release version that changes will be released in ahead of time.

Issue #373, opened 2020-02-19
clean_run: true does not work for sql statement


palin:
Using jdbc plugin 4.3.13
Using logstash in docker: docker.elastic.co/logstash/logstash:7.1.1

In theory **clean_run** setting of the jdbc plugin decides:
> Whether the previous run state should be preserved
(https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html#plugins-inputs-jdbc-clean_run)

Again, in theory if we set **clean_run: true**, the sql_last_value should become '1970-01-01 00:00:00'

> If clean_run is set to true, this value (sql_last_value) will be ignored and sql_last_value will be set to Jan 1, 1970
(https://www.elastic.co/guide/en/logstash/current/plugins-inputs-jdbc.html#_state)

However, in practice, the _clean_run: true_ in the example below still causes the pipeline to keep the last run state under the :sql_last_value parameter (the date of last run in our case).

*:sql_last_value* is set to '1970-01-01 00:00:00' only when the pipeline starts for the first time.
2nd and next runs don't start from '1970-01-01 00:00:00', the sql query still contains value of the last run.

Logs:

**1st run is OK:**
```
app_1 | [2020-02-19T11:33:01,735][WARN ][logstash.inputs.jdbc     ] (0.103191s)
app_1 |       SELECT
app_1 |         rg.id,
app_1 |         rg.name,
app_1 |         r.name AS retailer_name,
app_1 |         r.trkref
app_1 |       FROM retailer_groups rg
app_1 |       JOIN retailers r ON rg.id = r.retailer_group_id
app_1 |       WHERE r.review_mark_enabled = 1 and rg.updated_at >= '1970-01-01 00:00:00'
```

**2nd run NOT ok:**
```
app_1 | [2020-02-19T11:34:01,038][WARN ][logstash.inputs.jdbc     ] (0.049631s)
app_1 |       SELECT
app_1 |         rg.id,
app_1 |         rg.name,
app_1 |         r.name AS retailer_name,
app_1 |         r.trkref
app_1 |       FROM retailer_groups rg
app_1 |       JOIN retailers r ON rg.id = r.retailer_group_id
app_1 |       WHERE r.review_mark_enabled = 1 and rg.updated_at >= '2020-02-19 11:33:01'
```

**3rd run and any later on NOT ok:**
```
[...]

PR #374, opened 2020-02-19
Fix saving pipeline state using record_last_run setting

palin:
This fixes the issue with clean run: https://github.com/logstash-plugins/logstash-input-jdbc/issues/373

The state should not be written if the **record_last_run** setting is **false**.

### DIFF
--- a/lib/logstash/plugin_mixins/jdbc/jdbc.rb
+++ b/lib/logstash/plugin_mixins/jdbc/jdbc.rb
@@ -252,7 +252,7 @@ module LogStash  module PluginMixins module Jdbc
       rescue Sequel::DatabaseConnectionError, Sequel::DatabaseError, Java::JavaSql::SQLException => e
         @logger.warn("Exception when executing JDBC query", :exception => e)
       else
-        @value_tracker.set_value(sql_last_value)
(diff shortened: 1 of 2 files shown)


marius-mcp:
Why is this not merged? The issue persists to this day...

ghost:
https://github.com/logstash-plugins/logstash-integration-jdbc/issues/121#issuecomment-2042584611

**clean_run => true
record_last_run => true
last_run_metadata_path => "/..."**
seems to be the solution and that is why this was not merged.

s137:
> Why is this not merged? The issue persists to this day... We applied the patch manually and it works...

I don't think this patch will be merged here, because the plugin is now part of the jdbc-integration-plugin in [this](https://github.com/logstash-plugins/logstash-integration-jdbc) repository, where I created the same issue, when I stumbled upon it, but the solution is of course the same regardless. 

Honestly I just made another jdbc-input which uses a separate query without `:sql_last_value` to work around it. It's not ideal, but it does the job.

ghost:
@s137 commented also there. See my previous message from here for the link.

Update https://github.com/logstash-plugins/logstash-integration-jdbc/issues/121#issuecomment-2045396133

PR #376, opened 2020-03-03
Point contributors to integration source

karenzone:
Add move notice to doc source

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -2,6 +2,18 @@
 :type: input
 :default_codec: plain
 
+
+//////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


docs/index.asciidoc
```diff
@@ -2,6 +2,18 @@
 :type: input
 :default_codec: plain
 
+
+//////////////////////////////////////////////
+// JDBC Input Plugin Source Has Moved       //
+// ---------------------------------------- //
+// The JDBC Input Plugin is now a part of   //
+// the JDBC Integration.                   //
```
robbavey: Should that be 'JDBC Integration Plugin'?

PR #378, opened 2020-03-05
Fixed Test Failures

robbavey:
Tests were failing due to a change in the latest Postgres docker image
requiring a password to be set by default.
This commit also includes necessary changes to support devutils 2.0


robbavey:
Thanks @kares!

